PR #2, opened 2021-11-11
Added a Github build workflow

PaytonTheMartian:
Just adds a small workflow to automatically build the mod when someone pushes or makes a PR to the repo. You can also download the artifacts from the build, being the compiled mod. (From `build/libs/` after `./gradlew build`)


DaFuqs:
this is great, ty!

PR #3, opened 2021-12-15
Updated Github Action to work with MC 1.18.x

efalsett:
Minecraft 1.18 upwards uses Java 17, so the Github Actions needed a small update there. :)


DaFuqs:
👌

PR #4, opened 2021-12-15
Almost complete translation for de_DE

efalsett:
It's not quite everything, but that's just a huge file!

Still better than nothing I guess


DaFuqs:
A huge understatement. That is a LOT!
Also thanks for the adjustments in en_us :)

PR #39, opened 2022-03-20
Build a better cache for the CompactingChest

### LINKED ISSUE #38: [BUG] Compacting chest causes massive tps lagg
**Describe the bug**
On the AOF5 server I play we had big problems with tps lagg, so we used spark to run a profiler and one compacting chest had the impact of 1.6% of the whole server ticktime. Nearly all of this time is happening in the recipemanager call.
This is happening because the cache is forgotten after each non successful crafting request, meaning it needs to look through the recipes again. Also if you have for example more than the required compacting items of something non compactable in the first slots it will combine those slots multiple times and check if a valid recipe for them can be found multiple times, causing more lagg.

I've created a better, shared cache solution in a PR coming soon, so this issue is kinda just a documentation of this problem

**To Reproduce**
Load a datapack with a large amount of crafting recipes and slowly fill the compacting chest using a hopper with new, non compressable items, causing the compacting chest to look for valid recipes over all item slots.

**Expected behavior**
The server not to lagg

**Mod version**
This issue was found in 1.2.8-1.18.1-midgame9, but I could also verify it in the latest github commit.

justliliandev:
Also improved the reaction time to automatic filling of the chest by trying to craft every second.

Closes #38

### DIFF
--- a/src/main/java/de/dafuqs/spectrum/blocks/chests/CompactingChestBlockEntity.java
+++ b/src/main/java/de/dafuqs/spectrum/blocks/chests/CompactingChestBlockEntity.java
@@ -5,6 +5,7 @@ import de.dafuqs.spectrum.inventories.AutoCompactingInventory;
 import de.dafuqs.spectrum.inventories.CompactingChestScreenHandler;
 import de.dafuqs.spectrum.registries.SpectrumBlockEntityRegistry;
 import net.fabricmc.fabric.api.screenhandler.v1.ExtendedScreenHandlerFactory;
+import net.fabricmc.fabric.api.transfer.v1.item.ItemVariant;
 import net.minecraft.block.BlockState;
 import net.minecraft.entity.player.PlayerInventory;
 import net.minecraft.item.Item;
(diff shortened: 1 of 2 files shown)


DaFuqs:
Awesome! I will have a look at it

justliliandev:
If you have any question, don't hesitate to ping me on discord

src/main/java/de/dafuqs/spectrum/blocks/chests/CompactingChestBlockEntity.java
```diff
@@ -206,4 +224,29 @@ public void writeScreenOpeningData(ServerPlayerEntity player, PacketByteBuf buf)
...
+			if (o == null || getClass() != o.getClass()) return false;
+			InputItem that = (InputItem) o;
+			return item.equals(that.item) && Objects.equals(nbt, that.nbt);
+		}
+
+		@Override
+		public int hashCode() {
+			return Objects.hash(item, nbt);
+		}
+	}
```
BasiqueEvangelist: Why not use Fabric API's ItemVariant here?

justliliandev: I'm usually modding with forge and didn't knew about the existence of ItemVariant

justliliandev: I've updated the PR to use ItemVariant

DaFuqs:
looking good!
I will release an updated version as soon as I figured out the best way to handle the current Origins incompatibility

PR #57, opened 2022-05-23
Simplified Chinese Translation

Wudji:


DaFuqs (commented):
looking good on first sight! I will test ingame

PR #67, opened 2022-06-17
A couple changes

sudo-CuriousLad:
1. New tooltips (pog)
2. Nature's staff ink tooltip now exists (forgot to add logic for using ink, will do that in a bit)
3. Nature's staff is now correct spelling


DaFuqs (commented):
Looking good. The Nature's Staff name needs to be changed in the guidebook, too, but I can add that no problem.

PR #74, opened 2022-06-29
idfk man, my head hurts

sudo-CuriousLad:


SollyW:
I also don't know man, and my head also hurts

DaFuqs:
The trick is to hurt others instead (always ask for consent!) ( ͡~ ͜ʖ ͡°)

PR #89, opened 2022-07-19
Migrate the Bedrock Armor model to Vanilla instead of Geckolib

Noaaan:
If anything should be commented, let me know


DaFuqs:
IMO everything is very much self explanatory, ty!

PR #104, opened 2022-08-14
Big changes

sudo-CuriousLad:
Some of these commits are me being a fucktard and not knowing how to use git, btu shouldn't affect the code. Review carefully please, there may be major fuck ups


sudo-CuriousLad:
I need to do this :kekw:

PR #105, opened 2022-08-15
Even more porting work

sudo-CuriousLad:
We're almost there guys! Only one fucking change left :hehe:


DaFuqs:
Something tells me that should not be a PR into the 1.18.2 branch 😘

PR #106, opened 2022-08-15
Even moh portin work

sudo-CuriousLad:
Sheer fuckery by me I don't know why


DaFuqs:
Looking at this, it seems mostly 👌
One thing that catched my attention is `has_structure/ancient_ruins.json` referencing the `c` namespace in json, but `spectrum` in code. Pretty much sure this will act up.

PR #110, opened 2022-08-27
Fix multiplayer related issues

Noaaan:
Big thanks to @gliscowo for helping me debug these.
The issues fixed here are mostlyrelated to multiplayer, with one minor fix to the language of the guidebook (old language strings were still "manual" instead of "guidebook"). 

In regards to recipes an issue was fixed where the recipes were not being serialized correctly if the amount of dust in it was 0. This would cause REI to completely break. 

The multiblock loading did work in singleplayer, but due to the nature of the `SERVER_STARTING` lifecycle event, it would fire too early for the multiblocks to be registered properly. This has been moved to a Mixin into `ClientPlayConnectionEvents` so that the multiblocks are always synced on reload. As a side effect this could in theory mean you could make hotswappable multiblocks if you desire, as it is always synced when `/reload` is run. 


Noaaan:
We also replaced the `MinecraftServer` instance with the one already used in Owo, since it saves LOC, and also has the benefit of removing itself once the server is stopped.

DaFuqs:
neat

PR #112, opened 2022-09-03
Fix for chiseled onyx blocks for pedestal structures on the 1.18.2

Norazan:
Cherry-picked the commit from the 1.19 branch that fixes this, would be nice to also use chiseled onyx blocks for pedestal structures on 1.18.2 modpacks :)


DaFuqs:
dang, did I forget to pick this? Ty!

PR #120, opened 2022-10-04
Moving Bedrock Armor parts, feat cape and flap rendering

Noaaan:
Some actual rendering on that model. No longer a static brick. Looks about 80% better


DaFuqs:
♥

PR #145, opened 2022-12-28
HealthOverlay compatibility

NerjalNosk:
It's janky, it mainly aims to fix the incompatibility, but at least it works.
Couldn't make the Divinity effect display as much health as intended with the HealthOverlay renderer active, so that has been ignored when with that mod active, but all the rest works pretty well so far.

Also includes an improvement of the compatibility check mixin plugin, along with some mixin refactoring to comply to the new system.


DaFuqs:
Fixed a single thing regarding to how divinity interacts with HealthOverlay.
That said, mixin' into Health Overlay itself it a major power move and the Mixin plugin system is really nice, ty!

PR #146, opened 2022-12-29
Finised HealthOverlay compatibility

NerjalNosk:
Now with support for HealthOverlay's AbsorbtionOverHealth config option


DaFuqs:
this is a gross, ugly hack

I love it

Terrails:
I was just going through the issues on my repo after leaving it be for a long time and Spectrum compat was reported a few weeks back. Decided to check it out and here I am, witnessing this beautiful monster. If y'all are fine with it I am all for making an easy to access API or giving you these values via fabric's ObjectShare. 
I am fully aware of how some parts of health overlay are wonky.... been just way to busy to work on it. Just shoot me up with any issues you see (open an issue on my repo or something) and I will look into fixing them. I have also just pushed a plethora of changes I had done in the last few days, so I am sorry as they definitely break this compat that you wrote.

Terrails:
I have just pushed changes that should make all of this unnecessary. You can now tap into ObjectShare in order to get "healthoverlay:absorptionOverHealth" and while I was at it, since I saw #145 I also added a way to force HealthOverlay to use hardcore hearts if you set "healthoverlay:forceHardcoreHearts" to true (and then back to false when the divinity effect is gone). Hopefully this makes your lives easier :)

EDIT: Backported it to 1.19.2 too

DaFuqs:
much appreciated! I'll take a look at it

PR #147, opened 2023-01-04
Updated HealthOverlay compat to match its new config system

NerjalNosk:
Update 7.2.1 of HealthOverlay broke quite a few things with compatibility, so there it is, a compat update


DaFuqs (approved):
👌
thanks a lot!

PR #156, opened 2023-01-22
One more texture update before updating my base!

Jenny77:
This time, I was finally being able to continue my job again!
More pictures, indeed!


DaFuqs (commented):
That chiseled_polished_calcite 👌👌👌👌

PR #160, opened 2023-02-13
Improve Bedrock Armor model cape rendering

Noaaan:
Reduces the shoulder clipping, and does not disable the front flap from being rendered when another cape, or Elytra, is present.
If you disable cape rendering on a skin level you will disable both the cape and the front flap. 

image


DaFuqs (approved):
nice

DaFuqs:
🥇

PR #163, opened 2023-02-28
Yet another layer fix for Bedrock Armor

Noaaan:
The shoes were too small. This PR fixes an issue with outer layers on the legs & feet of the player.
Before:
image
After:
image


DaFuqs:
Das Boot

PR #164, opened 2023-02-28
Added decay pickup and block entity spread configs

Electro593:
This PR adds the ability to prevent picking up decay with bottles, as well as to configure whether it can spread to block entities.

Note that I was unsure whether the lack of ForfeitureDecayTickRate validation was a bug, so it's the same as it was before for now. Also, the non-en_us lang files need to be updated for the new options.


DaFuqs (approved):
Perfection. Looking good!

PR #165, opened 2023-02-28
Bonemeal for blood orchids and mermaid's brush

Electro593:
This PR makes bonemeal instantly grow blood orchids by one age tick, and mermaid's brush by several depending on certain conditions.


DaFuqs (approved):
👌

PR #169, opened 2023-03-05
Naturally Generated Dragon Fossils

Electro593:
This PR makes dragon fossils naturally generate, though frequency and shapes will probably need to be tweaked.


DaFuqs (commented):
I feel the DragonFossilPlacementModifier could be replaced with `minecraft:environment_scan`, but that does not include the y+15 offset yet, so I see where you are coming from.
The y+15 could result in a position out of height limit, that will need another check, everything else looks perfect to me.

PR #172, opened 2023-03-13
Full EMI integration

### LINKED ISSUE #111: [Feature] EMI support
Similar to REI, but with some handy features.
[https://www.curseforge.com/minecraft/mc-mods/emi](https://www.curseforge.com/minecraft/mc-mods/emi)

emilyploszaj:
Resolves #111 

This is a moderately large pull request that implements, as far as I can tell, full EMI integration for Spectrum. It is worth noting that I've not gotten the chance to play through Spectrum (though with EMI integration maybe I'll get the chance soon) so I would not be able to notice everything. Included is full recipe support with (to my best understanding) hiding of locked recipes, full tag translation or exclusion, workstations, and some extant cleanup related to my work.

I tried to maintain some widget construction patterns as shown in the REI plugin to keep it as familiar to you as I can, however, it's worth noting that the EMI integration is much smaller and I took other opportunities to simplify my implementation.

Here are some notes for maintenance:
* EMI has tag translations and models. This means tags can be given names (or hidden with another resource pack file) and optionally given models. I did not give any tags models for Spectrum, I'm not familiar enough to be able to do that, I don't think. EMI by default has "chimera" tag models showing multiple things at once, if you're interested see `#minecraft:coals`, `#minecraft:flowers`, and `#minecraft:beacon_payment_items` in game.
    * Keeping up with this can be as minimal as simply translation new tags you add, all tags that currently exist have been translated by me to the best of my ability.
    * Documentation for this system is [here](https://github.com/emilyploszaj/emi/wiki/Tag-Translation).
* Of course, adding new recipe types would mean adding new EMI recipe types. I believe I've provided adequate examples and the work required should be minimal, but if this is not the case, feel free to reach out to me for pointers.

Here are some changes I made otherwise:
* The buildscript now has the ability to pick between runtime recipe viewers in the properties.
* I adjusted a couple tag definitions to change the ordering for EMI's donor item display to make more sense.
* I removed your definition of the `minecraft:pressure_plates` item tag, this tag doesn't exist in vanilla and I couldn't see it anywhere in the codebase, so it was just a loose end.

Outstanding:
* There are 15 runtime warnings from EMI about tags that are not `c:` or `spectrum:` being untranslated. I don't believe this should be considered your responsibility, and in the future I want to alter how EMI displays myself this to avoid unavoidable spam like this.


DaFuqs:
The code is really concise, looks easy to maintain (people were not joking that it would be easier than REI) and a good reference point for all future integrations that I will add myself.

I will have to test a few cases that are not immedately obvious to me or might be missing (recipes with the `secret` flag to never be shown), but I will see how it plays out ingame and can get some minor adjustments done myself.

Thank you lots! May be that I will be bugging you with one or two questions, but you have been a giant help already

DaFuqs:
Looking great! 2 minor things that I noticed that you may be able to shed some light on:

1. Sometimes, single recipes, even in the same recipe type, do not render their gui backgrounds?
image

2. Some Spectrum recipes are flagged as secret in their json and can be queried in `recipe.isSecret()`. Those should never get displayed by recipe viewers. I tried overriding [hideCraftable()](https://github.com/DaFuqs/Spectrum/blob/c0b9377f8755cd2f6ebab9fd2c179a8385574b3e/src/main/java/de/dafuqs/spectrum/compat/emi/GatedSpectrumEmiRecipe.java#L33), but that method does not seem to be queried.
The [EMI Wiki entry on Recipe Filers](https://github.com/emilyploszaj/emi/wiki/Recipe-Filters) does not mention a way, besides duplicating the recipe ids in a separate file (double the work, error prone)

emilyploszaj:
1) Those output slots are rendering properly! The checkerboard pattern is an explicit texture used for inputs and outputs that have a chance component, you can see the chance tooltip when hovering.

2) If those recipes should never be displayed, you can easily just not tell EMI about them. I suggest adding an `if (!recipe.isSecret()) {`  check before you call `registry.addRecipe` in`SpectrumEmiPlugin.addAll()` where basically every recipe except the block conversion ones are added.

PR #173, opened 2023-03-13
Update TrinketChangeCriterion.java

chyzman:
fixed ur optionals. #IHateOptionals

### DIFF
--- a/src/main/java/de/dafuqs/spectrum/progression/advancement/TrinketChangeCriterion.java
+++ b/src/main/java/de/dafuqs/spectrum/progression/advancement/TrinketChangeCriterion.java
@@ -42,11 +42,11 @@ public class TrinketChangeCriterion extends AbstractCriterion<TrinketChangeCrite
 	
 	public void trigger(ServerPlayerEntity player) {
 		this.trigger(player, (conditions) -> {
-			Optional<TrinketComponent> trinketComponent = TrinketsApi.getTrinketComponent(player);
-			if (trinketComponent.isPresent()) {
+			Optional<TrinketComponent> trinketComponent = TrinketsApi.TRINKET_COMPONENT.getNullable(player);
+			if (trinketComponent != null) {
 				List<ItemStack> equippedStacks = new ArrayList<>();
 				int spectrumStacks = 0;
-				for (Pair<SlotReference, ItemStack> t : trinketComponent.get().getAllEquipped()) {
+				for (Pair<SlotReference, ItemStack> t : trinketComponent.getAllEquipped()) {
 					equippedStacks.add(t.getRight());
 					if (t.getRight().getItem() instanceof SpectrumTrinketItem) {
 						spectrumStacks++;


DaFuqs:
why

[The Trinkets API](https://github.com/emilyploszaj/trinkets/wiki/Trinkets-API) specifically calls for using that method instead of direct, raw access

chyzman:
api bad

gliscowo:
The actual reason why this change is an improvement is the following: the trinket component *cannot* be missing (and even if it were, that would certainly be a reason to crash since a core component of the mod is broken) and avoiding this optional by just directly using `get(...)` on the component key saves you the allocation and makes for cleaner code as you save yourself a redundant check and a level of nesting

Just giving my two cents (and I may or may not be horribly biased as I think Trinkets has *terrible* API), but of course this is your call to make

To further elaborate on why the component cannot be missing - it is registered for [all living entities and players](https://github.com/emilyploszaj/trinkets/blob/47b9692245081c35ecf93968b97de8f0aa857ff7/src/main/java/dev/emi/trinkets/TrinketsMain.java#L118), and is thus guaranteed by CCA to exist (otherwise the core premise of CCA would be, let's say, somewhat useless). 

Furthermore, as said above, even if the component somehow went missing, that definitely constitutes a crash condition since the entirety of Trinkets would be broken. There is no reason why the trinket component should be optional - all it does is incur an additional allocation and the responsibility of handling an Optional even though it is always populated

Please excuse this rant, but I have some pent-up frustration over this particular bit and a few other areas of Trinkets API

emilyploszaj:
In general going around an API's supported codepaths is just setting up brittle code. I don't anticipate this particular PR causing any issues (though I share the confusion for its motivation), but I did want to correct some assumptions about Trinkets' API design from glisco. 

Fundamentally, the Trinkets API is telling you that a living entity might not have a trinket component. Your assumption that this could only come from an error is incorrect, it could also be an implementation detail. Trinkets' API is explicitly telling you this may be the case. Allocation as a concern could prompt the Trinkets to not hook up a component for entities which do not have enabled slots (which, in most cases, is almost all of them). Assuming promises from the current implementation not promised to you by the API plagues areas all around Minecraft modding, and software engineering more broadly, causing non-breaking updates to crash users who assumed they knew better than the API they were interfacing with.

I'll concede you are right about optionals being stinky. When the modern API was designed by C4 and I it drew inspiration off of C4's experience with Curios' systems mixed with the more open ambitions of Trinkets and we had designed the hypothetical implementation case I just described, but by the time we wrote the implementation it was just easier to put components on everything and adjust it later if performance became a concern. It is often repeated that premature optimization is the root of all evil, this case is no exception. Were I to design this API modernly, I'd simply make the method nullable, because optionals have gone out of fashion.

Despite this, it's simply more beneficial to mod authors and users to not have breaking changes in Trinkets' API. The current systems are overengineered for certain features seldom used, and despite the 1.16 rewrite still holds on to some tech debt from the initial 1.14 implementation. But, it's featureful, unified, it works, and it's not going to require a rewrite every major update. That's really what matters at the end of the day to developers and users.

I would encourage the usage of the suggested paths in this case for Trinkets and in other cases in the future. As a rule, API is the promise, and impl is not.

gliscowo:
While I 100% agree with you on the fact that relying on implementation detail is a bad idea, in this specific case I don't think that goes to the core of my reasoning. The main point I was making is that if the Player's trinket component is missing, that 100% most definitely substitutes a reason to crash since, as I said in my previous comment, that means a core part of the mod is fundamentally broken.

And at the end of the day, I personally think that having breaking changes on the next major Minecraft release to fix some outstanding issues with the API (like the use of optionals, or there being no good way of equipping Trinkets for example) would majorly benefit everyone involved - I doubt anyone would mind an extra 10 minutes of porting effort if it means getting cleaner and easier to use API as a result

emilyploszaj:
Time permitting it might be worthwhile to add methods to the Trinkets API in a new major release, but your reasoning remains incorrect. The concept that a player must have a trinket component is simply untrue, and actively untrue by virtue of what the API explicitly specifies. I could add a command tomorrow that lets players toggle trinkets on and off for themselves potentially removing this component and it would require no API changes, unless people made unfounded assumptions.

DaFuqs:
^ this tbh

PR #174, opened 2023-03-14
Comparator support for chests

rhit-seilerar:
This PR adds comparator support for all Spectrum chests.

Note: There are a number of other blocks that might be worthwhile to add comparator support to. The placeable inventories that don't currently output a comparator signal are presents, bottomless bundles, color picker, crystal apothecary, potion workshop, and titration barrel. Also, maybe idols should emit some signal when they're triggered.


Electro593:
Whoops, wrong account.

DaFuqs:
Ty!
- bottomless bundles, color picker, crystal apothecary, potion workshop, and titration barrel would definitely make sense.
- Since people should not know what is in a present, I do not think they would need it, too.
- Idols should get a rework down the line anyways to make them more distinct and memorable

PR #175, opened 2023-03-15
Comparator support for all other Spectrum inventories

Electro593:
This PR adds comparator support for titration barrels, bottomless bundles, crystal apothecaries, potion workshops, and color pickers. It also makes the fluid content in titration barrels and fusion shrines contribute to the output.


DaFuqs (approved):
The Glow Vision Goggles could use an alternate tooltip once the player unlocked Ink, but that is a non-issue. Looking good!

PR #192, opened 2023-05-23
Added Ears compat

MayaqqDev:
UwU


DaFuqs:
Looks fine to me!
One thing though, I guess the torso renderer should also be cancelled when wearing other types of chestplates? (like gemstone)

DaFuqs:
I'll add that myself, then. Will be quick

PR #195, opened 2023-06-04
Dropped items add fluid to the fusion shrine

Electro593:
This pull requests allows dropped buckets, bottles, and other fluid-holding items to insert and remove fluids from the fusion shrine.


DaFuqs (commented):
looking good

PR #196, opened 2023-06-04
Fixed kelp-induced infinite black materia

Electro593:
This PR adds a quick fix for midnight solution creating infinite black materia when above kelp. This could be made to apply to other non-solid blocks as well, to prevent infinite materia generally, but that would take careful consideration to avoid destroying things like ender portals when materia lands on them.


DaFuqs:
Perfect!
I almost feel we could filter for Material.UNDERWATER_PLANT here, or make it a block tag instead (corals?), but these never have been an issue yet, with kelp being the biggest contender, being everywhere.

PR #200, opened 2023-06-17
stripped coloured log textures

800020h:


DaFuqs (approved):
looking good! Thank you a lot

PR #202, opened 2023-06-24
Allow Loot Tables to be used with Item Bowls and Item Roundel

Noaaan:
Technically also applies to any of the crafting stations, as they also extend the InWorldInteractionBlockEntity, which this has been implemented upon.

Based upon code from LootableContainerBlockEntity

Did you know that Hoppers can accept Loot Tables when placed? Crazy stuff.


DaFuqs:
Neat. Thanks a ton!

PR #207, opened 2023-07-12
Rename fiery/blizzard powder and malachite crystals

Electro593:
This pull request renames Fiery Powder -> Incandescent Essence, Blizzard Powder -> Frostbite Essence, and Malachite Crystal -> Refined Malachite.

Note that the zh_cn translations for these are now probably inaccurate.


DaFuqs:
Neat, ty!

PR #214, opened 2023-07-19
amphora bottom fix

800020h:
I didn't test, but it should work

@Azzyypaaras said to use the `light_top` textures


DaFuqs (approved):
Looks good to me. Ty!

PR #215, opened 2023-07-20
Fix Compatibility with Terralith

catter1:
According to the bot in your Discord server, Terralith modifies the terrain in such a way as to make two of the Spectrum's plants very rare. Although this is technically true, we discovered that it can be fixed easiest on Spectrum's end.

All placed features in the game referenced from biome files start at **y=0** in the corner of the chunk they start in. Placement modifiers such as `in_square` and `height_range` need to be used in order to place the features anywhere useful. After these  placement modifiers are ran, a `biome` placement modifier is meant to be run to make sure the feature didn't "leak" into another biome. However, Spectrum does this out of order. Instead of checking for a Swamp biome at the surface, the biome check is applied before the heightmap adjustment, so it checks at y=0 instead.

Now, why does this cause Quitoxic Reeds and Mermaids Brush to be rare with specifically *Terralith*? Terralith adds several cave biomes such as Andesite, Diorite, and Granite Caves, which commonly generate around y=0. Since the biome check is applied before the heightmap adjustment, the placed feature sees those caves instead of a Swamp, and fails. My commit(s) adjust this order so it will no longer happen.

Thank you for your time, and a *ton* of thanks to @Apollounknowndev for helping me work through this.


catter1:
Another note: one thing I've seen floating around in your DIscord is that Terralith modifies the swamp to make the clay deposits slightly smaller. This does seem to be true: when testing my fixes, although Quitoxic Reeds were spawning, they were still a bit rare. I will work on making an update to Terralith (1.18.2 for now) so clay deposits are bigger in swamps when Spectrum is loaded.

DaFuqs:
This goes a long way to explaining why worldgen worked after I moved the plant generation [to json](https://github.com/DaFuqs/Spectrum/blob/1.19-deeper-down/src/main/resources/data/spectrum/worldgen/placed_feature/quitoxic_reeds.json), as I changed the order there. That change is not in the public release yet, though.

Now that you mention it, what was I doing back then? To be fair, this was my first experience with worldgen, so I was probably young and stupid. Thanks to everyone involved!

As the plants are renewable once the player has found some, there is no need to jump through hoops to increase the clay rate. As long as you find them every few chunks, that is all well and good.

PR #241, opened 2023-08-13
Re: Restaurants Rapidly Resonance Relocating Ravaged Roasts*

800020h:
some of this PR is so they can do just that

_*should probably be called "resonance mod integrations"_

notes:
- I couldn't do `#minecraft:leaves` `[persistent]` because I don't have the technology yet
- I didn't do supplementaries signs and doormats because I didn't want to try doing `supplementaries:sign_post`, and didn't want inconsistency within one mod's integration
- I didn't do `nethersdelight:stuffed_hoglin` because it's weird
- I only went through my modpack project and AOF6
- it's entirely possible I missed something


800020h:
farmersdelight canvas signs are in `#minecraft:signs` the block tag, but not the item tag
so I'm not quite sure if they need separate handling with the new way of doing it

DaFuqs (approved):
👌👌👌

PR #266, opened 2023-10-21
Fix `smooth_basalt_slab` stonecutting ingredient

ConductiveFoam:
Seems wrong and an easy fix.

Sorry if I missed datagen, didn't see any at a glance. (Then again I don't know Fabric at all)


DaFuqs (approved):
👌

DaFuqs:
thank you!

PR #294, opened 2023-12-08
Fix Quilt related issues

EvieTheOwl:
Makes Spectrum work for Quilt 1.20.1. Just some minor things that need changing to get it working. One of them is to downgrade the dependency requirement but it's by one minor version and doesn't impact the mod at all.

- Downgrade dependency requirements (Quilt hasn't updated yet :/). The newer FAPI stuff wasn't used anyway so it's fine :3.
- Fix Criteria accessor issues. See https://github.com/DaFuqs/Revelationary/issues/9 for more details.


DaFuqs (approved):
Looking good. Ty!

PR #305, opened 2023-12-22
maxgamer134's contributions

maxgamer134:
_**THIS**_  is a real pull request. 
I want to save my work so that I don't lose it! I reformatted a large chunk of the new guidebook, and I also want others to see it.


src/main/resources/data/spectrum/modonomicon/books/guidebook/entries/decoration/banner_patterns.json
```diff
@@ -30,7 +30,7 @@
       "text": "Resources and tools you come across can be used directly in the Loom as patterns."
     },
     {
-      "type": "modonomicon:crafting",
+      "type": "modonomicon:crafting_recipe_recipe_recipe_recipe",
```
Electro593: I think this might have replaced a few too many times

maxgamer134: huh weird

PR #308, opened 2023-12-25
Patch Compatability With Sinytra Connector.

eehunter:
This removes the line of code that, according to the comment, won the "risky mixin of the day award". This fixes one of the issues that causes the game to crash when attempting to load Spectrum using Sinytra Connector.


eehunter:
The above changes _should_ allow Spectrum to run on Sinytra Connector without crashing. There shouldn't be any effect on the game when run using Fabric. There are still a few bugs when running it with Sinytra, though.

DaFuqs:
Awesome! I'll take a in-depths look at all your changes in a few days.

eehunter:
I also patched Revelationary to work with Sinytra Connector, but the code is a mess right now. Revelationary 1.3.6 doesn't crash or anything when used with Sinytra, but it renders cloaked items as missing models. I'll clean up the patch and submit a pull request there when I get the chance.

eehunter:
I made a mistake on the last commit yesterday, by forgetting to add two of the three changed files to git. That's fixed now. Also, as a heads up, when you review the changes, please pay special attention to the ExplosionMixin file. The changes I made allow the mixin to apply correctly on both Fabric and Connector, but I did not test whether or not it _behaves_ correctly, as I'm trying to avoid Deeper Down spoilers until I can play through the mod again.

eehunter:
Also, there is one remaining graphical bug I know of. The mixin that makes splash particles appear when an entity falls into one of Spectrum's custom fluids only works on Fabric. I had to make it disable itself when Connector is detected to prevent a crash due to an injection error. I don't think it will be too hard for me to make a Forge-compatible alternative.

eehunter:
After attempting to explode Dragonbone on my playtest world, I can confirm that the ExplosionMixin in this pull request does _not_ work on Forge. I have not tested it on Fabric. I will attempt to fix this issue ASAP, however I am traveling today and do not have access to my computer.

Noaaan:
Just a quick note - This PR seems to have expanded quite a bit in scope from just 'Remove "award-winning risky mixin" code'. I think a rename of this is in order, as it seems more to aim to rework the codebase to work on FML via Sinytra Connector

eehunter:
That is correct. When I initially made the pull request, I was not aware that further commits to my fork would be automatically added to it. I had been planning to make a separate pull request for the other stuff, but that turned out not to be necessary. The name has been updated accordingly.

eehunter:
I have fixed the ExplosionMixin on Forge. It should work correctly on Fabric as well, but this has not been tested. 
The last two remaining issues I am aware of are as follows:
The Universe Spyhole does not create an x-ray effect.
The Gemstone Lamps do create an x-ray effect.

DaFuqs (approved):
I will run a few tests, but all changes look perfectly fine to me and even cleaner than before in some cases. Thank you lots.

DaFuqs (approved):
I tested as much as I could and can confirm fabric is still running well and good.
Did not have time to look into sintra yet, but I am wure it will work out on Sintra's side, too. 🏅

DaFuqs:
Thank you lots!

eehunter:
Oops, there was another commit I forgot to push that prevents a crash with Sinytra on dedicated servers! I think I'm gonna open another pull request real quick.

PR #312, opened 2023-12-30
Jade Vine Petal Block > Jade Vine Petals

Aeonsteel:
Added a recipe for turning Jade Vine Petal Blocks back into Jade Vine Petals, as Effulgent Feathers (which has a similar set of blocks) has a recipe to reclaim the raw material from the block.


DaFuqs:
That should be a shapeless recipe, but otherwise looking perfect.

Aeonsteel:
> That should be a shapeless recipe, but otherwise looking perfect.

Okay, I edited it to be shapeless. Sorry for that!

DaFuqs (approved):
perfect

PR #314, opened 2024-01-05
Another Sinytra Connector fix.

eehunter:
There was a compatibility issue on dedicated servers that I forgot to commit and push the fix for before my last pull request was merged. The captured locals on one of the ExplosionMixin injections caused a crash on dedicated Forge servers, despite working on clients and singleplayer. This patch uses two other injectors to capture the necessary locals without causing an injection issue with modified bytecode. Unfortunately this also makes the connectormod/present version of the mixin a little less clean.


DaFuqs (approved):
Oh yeah, that should definitely get captured.

PR #315, opened 2024-01-06
Rod of submission

MayaqqDev:
UwU? I think would be very cool and useful but is on u!""!!


DaFuqs:
This is a weird af addition, but I love it.
Though I feel it should be in an addon mod cause it clashes with the original theme a bit. Adding the Stratine Plug, too. :tomfoolery:

DaFuqs:
=> https://github.com/DaFuqs/SpectralMemes :tomfoolery:

I also added all the assets for the Stratine Plug, if you still have the code.
I also remember people making a Bedrock Flail and oversized Malachite Shears, but I'll have yet to find the textures again.

DaFuqs:
Moved into Spectral Memes

PR #316, opened 2024-01-08
Made Upgrade Blocks render the little silly floating disks when there is a opaque block above them and not just air because yes! Also changed create:honey tags to c:honey.

MayaqqDev:
UwU


src/main/resources/data/spectrum/recipes/titration_barrel/bristle_mead_fluid.json
```diff
@@ -14,7 +14,7 @@
       "count": 2
     }
   ],
-  "fluid": "create:honey",
+  "fluid": "c:honey",
```
DaFuqs: Pretty sure the Titration Barrel does not support fluid tags right now. That would definitely be a nice addition for mod compat.

MayaqqDev: oh okay, my bad

DaFuqs:
Good stuff

PR #317, opened 2024-01-08
Added additional clarification that you can process stacks of ingredients in the fusion shrine

Gronglegrowth:


DaFuqs:
~~I feel that info would be best suited at the position the player looks up their first fusion recipe.
While they...~~ wait, that info is in the Onyx page already, is it? Yep.

Well, adding the info to the Fusion Shrine itself can't hurt.

PR #325, opened 2024-01-22
Fix Nature's Staff compatability issue with Sinytra Connector.

eehunter:
For some reason the Nature's Staff wasn't doing anything on Forge except beginning to play its sound effect repeatedly and immediately stopping each time. Setting the max use time fixed this, but only if it was set to a number that wasn't too high for some reason. I believe the upper bound is due to an incompatibility with another mod, but I don't know how to begin to go about diagnosing it.


DaFuqs:
weird. But works for me. Since the Radiance Staff has a similar way to use it, it may have a similar issue?

PR #326, opened 2024-01-24
Fluid input tags for titration barrel

imreallybadatnames:
Adds support for fluid tags using FluidInput
Tags are used just the same way as fluid identifiers: replacing a fluid identifier with a tag one(without the `#`) works just the same
NOTE: not yet (fully) tested, there may be certain issues that I either did not see or are inconspicuous. May need refactoring. Could also program the recipe deserialization to prefer searching for the tag first, in order to allow tags with the same name as a fluid identifier OR only parse IDs that start with `#` as tags.


imreallybadatnames:
apparently `c:honey` doesn't exist (at least in my case)

imreallybadatnames:
According to my testing, tags work just fine, however Spectrum still complains about the ID missing. Strange...

imreallybadatnames:
talking about [this line](https://github.com/DaFuqs/Spectrum/pull/326/commits/3bccf6ee75c2e0b002acc5c3e94720f6cf639365#diff-9e3345a2a8dbb382a34d8f157aae9ce68e79d4af8a464557b7a5468070588821R46)

imreallybadatnames:
> According to my testing, tags work just fine, however Spectrum still complains about the ID missing. Strange...

Apparently, the tags somehow do not exist during the time the recipe is deserialized. Although I did see that the registry *had* tags once. The tags are really inconsistent on my end.

DaFuqs:
Do you have mods present that add that tag? We could also ship that tag with optional honey entries for both Create and Bumblezone ourselves.

imreallybadatnames:
I've had Create for purposes of testing, however it did not add the `c:honey` tag.
Turns out I've used a version of Bumblezone that my minecraft instance did not register, and after adding it in, Spectrum still complained about the lack of such tag. At least it shows up in EMI properly.
After further testing, it indeed appears that the `FluidInput` has been set to `EMPTY` due to this, however, EMI still shows the recipe without error. That is peculiar.

imreallybadatnames:
The recipe too appears to work just fine.
I have successfully brewed bristle mead(whose fluid tag I substituted for `c:honey` in the titration barrel recipe), and it worked flawlessly(albeit I did not wait a whole day just to test this, I simply used the paintbrush trick to get it done quickly)

imreallybadatnames:
Also tested brewing with Bumblezones royal jelly, which also appears in the `c:honey` tag. That worked too.

imreallybadatnames:
Now I have to figure out just *how* and *why* it still works(with the `FluidInput` being empty yet somehow ending up correct), and why the registry seemingly doesn't have any tags while deserializing the recipes.

imreallybadatnames:
My only guess is that the tags are somehow populated after the recipes have been deserialized(not yet confirmed), but that doesn't explain the fact that the recipe somehow corrects itself.

imreallybadatnames:
Ookay, I just noticed the recipes being loaded twice, the first time being *before* I agreed on the experimental settings, in which the tag thing parsed as `EMPTY`, and the second time being *right* after I did so, with the tagged `FluidInput` actually working. So this is pretty much a non-issue that I erroneously observed as being an issue.
However, the tag parsing mechanism is yet to be discussed, i.e. Should tags be preferred over individual fluid IDs in terms of parse order (since there exist tags of the same name as fluid IDs which serve the purpose of compatibility)? Or should I separate tags and fluid IDs altogether by interpreting `#` as a tag ID specifically? I should note that tag IDs with the `#` **cannot** be parsed by the Identifier-parsing functions.

imreallybadatnames:
An alternative option for this would be to either reformat the `fluid` field in the titration barrel recipes(which would require changing them all) or introduce a separate mutually exclusive field of e.g. `fluidTag`

imreallybadatnames:
Tested with REI, only problem I found is that the bucket ingredients are double in the case of honey. Going to look into that.
image

imreallybadatnames:
Turns out the problem is that REI thinks that there are double the entries in the tag than there actually are. Weird.

imreallybadatnames:
Nope, not weird, the actual entries are double because there are both still and flowing fluids. Nevermind.

imreallybadatnames:
Figured I'd just use some `Set` type from fastutil to deal with bucket deduplication. Testing.

imreallybadatnames:
Oops, forgot to send an update. Works like a charm.
image

imreallybadatnames:
Everything seems to be normal. Waiting for tag parsing discussion.

imreallybadatnames:
Also, one more thing: Should the EMI titration barrel recipe fluid input behavior be mirrored into REI? It seems inconsistent to have one mod display fluid buckets and the other display the fluids themselves. Changing the REI recipe to display the fluids directly would also better represent the accepted fluid inputs in the case of using automated fluid transfer.

imreallybadatnames:
... What?

imreallybadatnames:
Reverted the strange occurence. Pushing normal commit now.

DaFuqs:
> Also, one more thing: Should the EMI titration barrel recipe fluid input behavior be mirrored into REI? It seems inconsistent to have one mod display fluid buckets and the other display the fluids themselves. Changing the REI recipe to display the fluids directly would also better represent the accepted fluid inputs in the case of using automated fluid transfer.

Yeah, showing up as fluid instead of a bucket, while less immersive, is definitely the more intuitive choice.

imreallybadatnames:
image
Now displays fluids.

imreallybadatnames:
I do have to note that the tag parsing question still stands.
Currently, there are 3 options:

1. Keep it as-is: Prefer fluid identifiers over tags of the same identifier. Makes usage of certain fluids explicit. Has no issues with the `c` namespace, as it contains only tags.
2. Prefer tag identifiers over fluids: Allows other mods to add fluid compatibility, but at the same time, these same mods add fluids that do not make sense for such recipes(e.g. having milk, chocolate and honey in the `#minecraft:water` tag).
3. Separate fluid and tag identifiers altogether by parsing identifiers starting with `#` as tags. INCOMPATIBLE with Identifier (de)serializing functions as `#` is an illegal character. However, (I think) it wouldn't be much work adding the extra logic in `TitrationBarrelRecipeSerializer`.

imreallybadatnames:
If (3) is chosen, I might as well bring the (de)serializiation logic into the `FluidInput` class itself to abstract the whole process. This is most important for (3), as the process would get more complicated, and the boilerplate code would get noticeably big if it were to be put in multiple places throughout `TitrationBarrelRecipeSerializer`.

imreallybadatnames:
I consider it really important to answer [this question](https://github.com/DaFuqs/Spectrum/pull/326#issuecomment-1912901240) before doing anything extra.

DaFuqs:
I feel it would be best going the vanilla route and handling the fluid input like vanilla does ingredients, basically a fluid ingredient, where it is explicitly stated if it should be interpreted as a fluid, or a fluid tag.

```json
"fluid": {
    "fluid": "bumblezone:liquid_honey"
}
```
vs.
```json
"fluid": {
    "tag": "c:honey"
}
```

Maybe rename FluidInput to FluidIngredient to match vanillas ingredient, which also matches items and item tags?

imreallybadatnames:
Should I just redo the (de)serialization and have `FluidIngredient` be something like a lightweight `Ingredient`?

imreallybadatnames:
I don't think matching the vanilla behavior down to the T would be optimal, and I think it would very much suffice to have a fluid input that only supports one item or tag entry(which also plays really well with mods like REI and EMI). Having to support multiple entries would bloat the handling code exponentially, in my opinion.

DaFuqs:
> I don't think matching the vanilla behavior down to the T would be optimal, and I think it would very much suffice to have a fluid input that only supports one item or tag entry.

yes, absolutely. Vanilla ingredients can match multiple items, which I do not see much use for for fluids (since those cases should use tags in the first place).

> Should I just redo the (de)serialization and have `FluidIngredient` be something like a lightweight `Ingredient`?

I feel like that would make it even more intuitive to use, so while not required it would be cool

imreallybadatnames:
@DaFuqs Just did the thing, please take a look.

DaFuqs (approved):
Looks perfect to me

DaFuqs:
Awesome. I'll play around with it a bit in my test env and then merge it 👌

imreallybadatnames:
@DaFuqs That's great! Anyway, I did not convert the titration recipes to the new fluid ingredient syntax, are you currently doing that or should I make the changes?

DaFuqs:
I will take care of those when testing.

imreallybadatnames:
Alright. I'll be waiting for the end results!

DaFuqs:
Looks like both Bumblezone and Create add their fluids to c:honey already, nice

DaFuqs:
Awesome work. Thank you a lot

PR #330, opened 2024-01-27
Add EMI recipe handlers for Spectrum recipes

Abbie5:


DaFuqs:
Tested and seems to work perfecty. Thank you!

PR #331, opened 2024-01-28
Add REI transfer handlers for Spectrum recipes

imreallybadatnames:
Complementary to PR #330
Self explanatory, adds "Move Items" support for some Spectrum workstations on REI.
Matches functionality added by the EMI PR.

### DIFF
--- a/src/main/java/de/dafuqs/spectrum/compat/REI/REIClientIntegration.java
+++ b/src/main/java/de/dafuqs/spectrum/compat/REI/REIClientIntegration.java
@@ -37,6 +37,9 @@ import net.fabricmc.api.*;
 import net.minecraft.block.*;
 import net.minecraft.screen.ScreenHandler;
 
+import java.util.List;
+import java.util.stream.Stream;
+
 @Environment(EnvType.CLIENT)
 public class REIClientIntegration implements REIClientPlugin {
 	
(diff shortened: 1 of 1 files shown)


imreallybadatnames:
Tested both Spectrum and vanilla cinderhearth recipes. Works without issue.
Tested both vanilla and Spectrum pedestal recipes. 1 issue: REI appears to ignore the gemstone powder proportions and simply puts 1 powder/craft instead of the needed amount.
Potion workshop brewing and crafting works without issue.

imreallybadatnames:
EMI does not have any issue with powder proportions, and I reckon it'd be a headache to dig into REI internals to try to fix that.

imreallybadatnames:
It appears that [REI takes specifically 1 item at a time](https://github.com/shedaniel/RoughlyEnoughItems/blob/14.x-1.20.4/runtime/src/main/java/me/shedaniel/rei/impl/common/transfer/InputSlotCrafter.java#L113) when filling item slots.

imreallybadatnames:
EMIs recipe filler [takes a different approach](https://github.com/emilyploszaj/emi/blob/1.20.4/xplat/src/main/java/dev/emi/emi/registry/EmiRecipeFiller.java#L286) and applies the item moves directly, while accounting for actual ingredient counts instead of the amount of possible crafts, which yields a correct result, and is also less confusing of a mechanism to analyze.

imreallybadatnames:
Forgot to mention that EMIs server-side mechanism for this [does the exact same thing](https://github.com/emilyploszaj/emi/blob/1.20.4/xplat/src/main/java/dev/emi/emi/network/FillRecipeC2SPacket.java#L79) but without the slot click spam.

imreallybadatnames:
However, EMIs mechanism relies on the input stacks being precomputed, with the computation being done by [`EmiRecipeFiller.getStacks`](https://github.com/emilyploszaj/emi/blob/1.20.4/xplat/src/main/java/dev/emi/emi/registry/EmiRecipeFiller.java#L133), and it is... rather large, to say the least. REI does not have anything near this, so it probably won't ever support proper item ratios for a long while. At the very least, the transfer handlers are still fully functional for recipe types other than Spectrum pedestal recipes, with the exception of a couple Potion Workshop crafting recipes, and is mostly functional otherwise.

DaFuqs:
That REI only handles single items is a bit unfortunate. Especially since Pedestal recipes can also use ingredient counts > 1 (albeit it's not used anywhere yet).
I agree with your research - there does not seem to be a good way to handle those cases.

How about excluding the gemstone powder slots from the quick transfer? The player will want to have these slots stacked at all cases, anyways.

imreallybadatnames:
I could try to make it such that REI will only complain about missing gemstone powders in the appropriate slots, not move them automatically.

DaFuqs:
Sounds like the best we can get 👌

imreallybadatnames:
Tested, works flawlessly.

DaFuqs:
Awesome, thanks!

Issue #310, opened 2023-12-29
[BUG] Nature's Staff forcing nearby players to right click


ChiefBjorn:
**Describe the bug**
When using Nature's Staff, nearby players unwillingly use right click.

**To Reproduce**
Use Nature's Staff, while a nearby player(within LOS, and ~40 block radius) is: holding a block/hoe/shovel/seeds/Radiance Staff/Paintbrush, or looking at a chest/levers/buttons/repeaters.

**Expected behavior**
While using Nature's Staff, the players nearby unwillingly: place blocks/seeds, till/flatten land, place lights, re-paint colored blocks(each depending where applicable), or force open chests, actuate redstone components.

**Mod version**
Spectrum 1.7.1 - AOF7

DaFuqs:
Fix by @imreallybadatnames will be included in the next bugfix-release, thank you!

PR #332, opened 2024-01-28
Fix weird staff interaction

imreallybadatnames:
The bug in question happens because `LivingEntity.usageTick` is executed on the server and ALL clients who have said entity loaded.
In order to fix this issue, a check must be implemented to ensure that the`clientUsageTick` method only executes on the player that is using the staff.
In my testing, a simple object identity check worked flawlessly for both the hosting player and a connected client instance.


imreallybadatnames:
Didn't realize issue # don't work in titles.
This PR fixes issue #310

DaFuqs (approved):
Clean, simple, spot-on.

PR #335, opened 2024-01-29
Remove jitpacked deps in fabric.mod.json

Mysticpasta1:
Jitpacked deps are made within the mod itself, adding fabric.mod.json deps means that those are completely seperate mods/libraries to download on CF or modrinth.


DaFuqs:
Hu. I really wonder why that is necessary. IT works perfectly fine on fabric, quilt and most of the Sintrya users.

PR #336, opened 2024-01-30
Add Enchantment Descriptions compat

Abbie5:


Noaaan:
Annoyingly running ED and Bookshelf in dev crashes due to a mapping error, but testing this in a prod environment seems to work. Thanks for the PR!

PR #337, opened 2024-01-31
Use Fabric Resource Loader API's built-in Programmer Art resource pack support for Spectrum's Programmer Art resource pack

Abbie5:
This means that instead of being a separate resource pack that players must apply if they want to use programmer art textures, Spectrum's Programmer Art resource pack will be activated automatically and invisibly when Vanilla's Programmer Art resource pack is selected.


KatTheFox:
I'm not sure this is a positive change- I, personally, often use Minecraft's programmer art, but like the new spectrum textures. Taking away that possibility for very little benefit (simplification of the resource pack menu? it's just one more pack, though) sounds like a bad idea.

Abbie5:
That's understandable, and I wouldn't mind if this PR gets rejected for that reason. I'll wait for some more feedback before closing though.

DaFuqs:
Hu, I did not know that is possible using FAPI now. In old versions, Spectrum used the lib "Programmer Art Injector" to achive this.
Personally, I see the point of wanting to enable the textures separately.

More opinions on the matter?

DaFuqs:
Thinking about it I agree with Kat here. While it's an additional entry in the resource packs section, I can see the benefit of being able to enable both Vanilla Programmer Art and Spectrum Programmer art separately

PR #340, opened 2024-02-02
Outcubus Core

Noaaan:
Title says it all. This PR aims to replace Incubus Core, as it is not really maintained anymore. 

To achieve this the following changes need to be made:
- Matchbooks need to be replaced. This is done with the new Matchbooks library: https://github.com/Noaaan/Matchbooks/
- Some code needs to be ported over. Notably the BlockLikeEntities, which are used for a lot of the moving blocks, as well as the custom contributor capes.


DaFuqs:
Perfect. I will do an in-depth review tomorrow.

DaFuqs:
It shall be done

PR #342, opened 2024-02-04
ends_delight resonance

800020h:
as far as I can tell this should work
but `uses_bites.json` didn't in my testing as I mentioned in discord
https://discord.com/channels/906264467636109333/906274936417554464/1203689412232220672

this should also be for 1.19.2


DaFuqs (approved):
👌

PR #343, opened 2024-02-04
Add idwtialsimmoedm compat

Abbie5:


DaFuqs (approved):
Will run a few tests in a real pack, but looks good to me already.

gliscowo:
One thing I'd add here is that the callback should probably return `null` when the enchantment description doesn't need obfuscating - simply delegating to the default handler makes much more sense in that case imo

DaFuqs:
done

PR #347, opened 2024-02-12
Update ae2.json: the block id of quartz_growth_accelerator now are growth_accelerator

luo-zi:


luo-zi:
Oh, I'm getting the submission to the wrong target

PR #348, opened 2024-02-12
Partially fix a minor incompatibility between Spectrum and Forge.

eehunter:
Forge, like the Milk mod, makes Milk a fluid. When Sinytra Connector is detected, the recipe for Clotted Cream should reflect this. Unforunately, due to an issue with Forgified Fabric API, right-clicking on a Titration Barrel with a Milk Bucket does nothing. This means that liquid milk must be piped in (I have not tested if this is actually possible) or inserted via the /data command. I do not know if there is a correct way to fix this on Spectrum's end.


DaFuqs:
Looks good to me.
One thing is still todo, which I can add myself: Updating the guidebook to reference the additional & renamed recipe if they are loaded.

DaFuqs:
Even better: I added the fluid tag c:milk and told the recipes to use that one, similar to the honey fluid ones. Seems to work just fine.

Issue #288, opened 2023-12-06
[BUG] Guidebook loading error


Cataratas:
When opening the guidebook, there's a loading error that says:

Error building entry spectrum:mod_integration/ae2 of book spectrum:guidebook
Error while building entry spectrum:mod_integration/ae2 page 1 of book spectrum:guidebook
Unkown item ID: ae2:quartz_growth_accelerator

Perhaps you meant "ae2:growth_accelerator"?

DaFuqs:
Looks like that block was renamed in 1.20, good catch!

DaFuqs:
Fixed for release tomorrow

luo-zi:
same problem in zh_cn folder

PR #349, opened 2024-02-13
Update ae2.json

luo-zi:
Sync from the English version #288


DaFuqs (approved):
Perfect, thank you.
While I checked if the guidebook loads, I totally missed to check the mod integration chapters, since they only load if the required mod is present.

PR #354, opened 2024-02-22
Custom item render

imreallybadatnames:
Adds the exact same thing as the previous PR, albeit with a different implementation.
Functionality is the exact same, and it works on the dedicated server now.


DaFuqs:
Looks good to me! Pretty performant and easy to use.
Looking at FAPIs DynamicItemRenderer (which I haven't looked into yet, to be honest) is there anything we need the additional functionality of your version for?

imreallybadatnames:
One of the most important information that FAPI leaves out, due to the location of its mixin, is the `leftHanded` boolean, which allows to properly rotate the rotated item (that will otherwise have wonky rotation if the uninformed renderer tries to apply the right-handed rotation in a left-handed position)
Even though it's more than possible to *still* account for this by doing careful matrix manipulation, it's just not worth the added complexity.

DaFuqs:
Looks good, works like a charm, perfect.

PR #355, opened 2024-02-22
fix(guidebook): clarify Exuberance description

KatTheFox:
as per https://discord.com/channels/906264467636109333/911958718067929118/1210029264733343785, Exuberance's guidebook description is currently too narrow (it specifies that it increases dropped XP from mob kills, while really it increases drops from a variety of sources).


DaFuqs (approved):
I like

PR #376, opened 2024-04-06
Fixed obvious errors in Russian localization

osiris-plus:
Fixed some of the glaring errors I could see.
Also I had to play around while translating "CMY ON" advancement title, so that people could understand it better.


DaFuqs:
Ty!

PR #384, opened 2024-04-22
Made the commands more command, is this what OOP means?

MayaqqDev:
Very cool, I realize your import system is weird but im pretty sure you can fix that so here it is! Maybe also some formatting, but it should be fine.


DaFuqs (approved):
Oooooh, I like that. I will test it out ingame and then give it a merger

PR #392, opened 2024-05-11
Fix Bedrock Anvil text field focus

imreallybadatnames:
Title. Also has a couple of nit changes, such as removing a couple redundant `setFocused` calls and handling screen closure properly.


DaFuqs (approved):
.✅

Issue #334, opened 2024-01-29
[Feature] Allow Construction and Exchanging Staves to pull from Bottomless Bundle in inventory


Abbie5:
Same as #278 but for Spectrum's own Bottomless Bundle.

DaFuqs:
Implemented in #398

PR #398, opened 2024-05-24
Allow staves to draw from bottomless bundles #334

KatTheFox:


DaFuqs:
Looking good!
Can you change the dedicated check for the Bottomless Bundle for a more generic one that can also handle vanilla Bundles, Shulker Boxes (and other modded stuff, maybe?).

KatTheFox:
Oh, good point, I'll look into that tomorrow

DaFuqs:
No hurry whatsoever

unilock:
> Can you change the dedicated check for the Bottomless Bundle for a more generic one that can also handle vanilla Bundles, Shulker Boxes (and other modded stuff, maybe?).

I think the intended way to do that is with the [Fabric API Lookup API](https://github.com/FabricMC/fabric/blob/1.20.1/fabric-api-lookup-api-v1/README.md); in this case, specifically [ItemApiLookup](https://github.com/FabricMC/fabric/blob/1.20.1/fabric-api-lookup-api-v1/src/main/java/net/fabricmc/fabric/api/lookup/v1/item/ItemApiLookup.java).

I'm not sure if there's actually a standardized item storage API *to* lookup, however...

DaFuqs:
> however...

1.20.6 Components in the corner be like:

KatTheFox:
I can at least allow it to draw from bundles. I'm not actually sure how shulkers store their items, since there's no documentation on the wiki (besides the new components which should of course make this way easier, however as of right now we're working with 1.20.1 so jank time it is), I'll do some testing. Bundles seem to just store their items in the 'items' nbt tag, which shouldn't be too hard to handle.

DaFuqs (commented):
I made a generic ItemProvider interface where items can be registered to be containers.

Botania, for example, does something similar using FAPI lookup, but there indeed does not seem to be a generic way to do this.
Components will make this quite a bit easier.

KatTheFox:
oh wow, nice. i was struggling to find a sensible way to do that

unilock:
> Botania, for example, does something similar using FAPI lookup, but there indeed does not seem to be a generic way to do this.

I made a basic mod that implements the Fabric Lookup API for finding `Storage<ItemVariant>` instances from `ItemStack` instances, which registers Bundles (untested) and Shulker Boxes (tested, working!) by default: https://github.com/unilock/ItemStorageProviderAPI

Notably, it supports extracting *and* inserting items to/from the inventory-bearing items.

Though I'm not sure it's worth finishing when it was already made obsolete in 1.20.5 :P

Noaaan:
The implementation might be obsolete, but the idea is still sound. 
In a 1.20.5+ context: How can you ensure that a custom component provided by a mod actually provides some form of storage? It could be as simple as iterating over them and checking whether they have an interface, but as it is right now this is not something a vanilla ContainerComponent has. 
For item-exclusive storage people might stick to the Container component, after all, it supports up to 256 slots. But for mixed storages and custom component implementations, an API like this would still be useful

PR #399, opened 2024-05-24
Clarify quitoxic reed biome

KatTheFox:
Because some people don't read the page that says "visit a swamp or mangrove swamp" and then complain that the next page says "(mangrove) swamp" and therefore they thought swamps wouldn't work. Apparently.


DaFuqs (approved):
✅

PR #400, opened 2024-05-24
Clarify how to place fading

KatTheFox:
Because apparently 'right click to place' is obscure. For the low low price of 16 shimmerstone, you too can learn basic minecraft mechanics.


DaFuqs (approved):
Definitely does not hurt.

PR #401, opened 2024-05-25
Update hummingstone hymn reqs to fix #395

### LINKED ISSUE #395: [BUG] Resonant Shards are only revealed if you've obtained hummingstone, but you can trigger the Hymn using natural generation
**Describe the bug**
The advancement for triggering the Hummingstone Hymn, which also serves to reveal the Resonance Shards it drops, requires you to have the advancement for collecting hummingstone as a prerequisite. Presumably it's set up this way because the assumption is that you'd need to pick up a bunch of hummingstone and put it together to get enough resonance for the Hymn. However, it's actually possible to trigger the Hymn using only a natural hummingstone formation, and if you do so before ever obtaining hummingstone blocks, the dropped Resonance Shards will be shrouded and unusable.

**To Reproduce**
- Find a particularly large natural hummingstone formation
- Trigger the vibration effect
- With enough hummingstone blocks, it'll all shatter and drop a bunch of shards
- These shards will remain shrouded until you go mine a different hummingstone formation and obtain hummingstone blocks

**Expected behavior**
It would make a bit more sense if the only criteria for revealing the Resonance Shards was having triggered the Hymn. Removing the prerequisite here doesn't seem like it would allow for any new sequence breaks.

**Minecraft version**
1.20.1

**Mod version**
1.7.11

**Screenshots**
image

KatTheFox:
Now just requires you to progress to the DD and do the actual hummingstone hymn

### DIFF
--- a/src/main/resources/data/spectrum/advancements/lategame/collect_hummingstone.json
+++ b/src/main/resources/data/spectrum/advancements/lategame/collect_hummingstone.json
@@ -23,11 +23,24 @@
         ]
       }
     },
+    "struck_up_hummingstone_hymn": {
+      "trigger": "spectrum:hummingstone_hymn",
+      "conditions": {}
+    },
     "gotten_previous": {
       "trigger": "revelationary:advancement_gotten",
       "conditions": {
         "advancement_identifier": "spectrum:lategame/spectrum_lategame"
       }
     }
-  }
+  },
+  "requirements":[
+    [
+      "has_hummingstone",
+      "struck_up_hummingstone_hymn"
+    ],
+    [
+      "gotten_previous"
+    ]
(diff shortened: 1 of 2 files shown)


DaFuqs:
Exactly the change I would have suggested for that one, too. Perfect.
One thing to note, though: This change will make the `strike_up_hummingstone_hymn` able to be unlocked without having its parent `collect_hummingstone` unlocked, too, which will feel weird from a players perspective and will leave some recipes and guidebook entries still locked.
I suggest adding striking a hymn as a second, alternative unlock condition for `collect_hummingstone`.

DaFuqs (approved):
Yeees. that is nice

PR #412, opened 2024-06-07
Added ru_ru lang for text and patchouli book

Kisuny:
Added translation of the patchouli book and a new translation from version 1.20 with the return of some modified keys from the new version


DaFuqs (approved):
I do not understand a single word, but I trust you with that one.
This is a HUGE contribution. Thank you a lot. The translation must have taken ages.

Kisuny:
It has taken a looooot of time, I will also be fixing the translation as I find problems, at the moment the only problems that are present in the translation are the text going outside the book, and that sucks

PR #416, opened 2024-06-16
head alteration recipes

KatTheFox:
adds head alteration recipes to the spirit instiller, namely:
recoloring for sheep, parrot, shulker, and axolotl heads
zombifying villagers, piglins, and hoglins
and creation of witch, mule, trader llama, glow squid, mooshroom, and wither skeleton heads


DaFuqs (commented):
Awesome!

Some minor stuff I noticed:
1. Some recipes use 1 and some 4 Vegetal. That should probably be consistent
2. All recipe unlocks implicitly check, if the crafting station required (in this case the Spirit Instiller) is unlocked, too. So the unlock advancement can be shortened a bit.

KatTheFox:
1. That was on purpose, since just recoloring a mob is a simpler application that imo would require less power- you can paint living mobs for a pittance of ink, so requiring 4 of pigment/vegetal to paint a head seems overly expensive
2. Ah, that makes sense, in that case I can remove the hidden advancement entirely and just make it depend on the treasure hunter head collection advancement

DaFuqs:
1. very reasonable, agreed#
2. thank you!

DaFuqs (approved):

PR #417, opened 2024-06-17
Update it to sync with the English version

luo-zi:
This fixed a loading bug where guildbook would report an error when the language was set to Chinese and there was also an Alloy Forgery


DaFuqs (approved):
thank you!

PR #418, opened 2024-06-18
Make Dragonrot Orange Again

imreallybadatnames:
Fixes the stupid color format bug in Sodium for Spectrum fluids by adding color provider overrides with proper logic.


DaFuqs (approved):
I do not like it that we need this, yet here we are.

PR #423, opened 2024-06-22
Fix Moonstone Pedestal Structure Name In Guidebook

KatTheFox:
The advancement calls the moonstone-tier pedestal structure the "Spectrum Complex", while the guidebook pages reuse the Onyx-tier name ("Spectrum Palace"). This is almost certainly an oversight,


DaFuqs (approved):
True, true

PR #427, opened 2024-06-27
Minor fluid collision refactor

imreallybadatnames:
Genericized neighbor update behavior in favor of fluids that only do fluid collision. The only behavior override is for midnight solution, which spreads black materia to blocks in addition to fluid collision.

The refactor was done with mod compatibility purposes in mind, since I have noticed that certain mods have extendable fluid collision handlers which handle fluid collision themselves (for e.g. collision simulation).

### DIFF
--- a/src/main/java/de/dafuqs/spectrum/blocks/fluid/DragonrotFluidBlock.java
+++ b/src/main/java/de/dafuqs/spectrum/blocks/fluid/DragonrotFluidBlock.java
@@ -4,6 +4,7 @@ import de.dafuqs.spectrum.particle.*;
 import de.dafuqs.spectrum.registries.*;
 import net.minecraft.block.*;
 import net.minecraft.entity.ai.pathing.*;
+import net.minecraft.fluid.FluidState;
 import net.minecraft.particle.*;
 import net.minecraft.registry.tag.*;
 import net.minecraft.util.*;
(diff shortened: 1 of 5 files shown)


imreallybadatnames:
I have to note that the actual mechanism may change slightly depending on requirements for aforementioned mod compatibility.

src/main/java/de/dafuqs/spectrum/blocks/fluid/DragonrotFluidBlock.java
```diff
@@ -39,43 +40,20 @@ public void randomDisplayTick(BlockState state, World world, BlockPos pos, Rando
...
-				world.setBlockState(pos, SpectrumBlocks.ROTTEN_GROUND.getDefaultState());
-				this.playExtinguishSound(world, pos);
-				return false;
-			} else if (world.getFluidState(blockPos).isIn(SpectrumFluidTags.MIDNIGHT_SOLUTION)) {
-				world.setBlockState(pos, SpectrumBlocks.BLACK_SLUDGE.getDefaultState());
-				this.playExtinguishSound(world, pos);
-				return false;
-			}
+
+	public BlockState handleFluidCollision(World world, FluidState state, FluidState otherState) {
```
DaFuqs: Methods that may return null should be annotated as @Nullable.

Looking fine, otherwise

DaFuqs (approved):
perfect, ty!

PR #429, opened 2024-06-27
Create mod fluid collision compat

imreallybadatnames:
Something for today. May do more fluid collision compat later.


DaFuqs (approved):
tested and seems to work perfectly. Ty!

PR #431, opened 2024-07-04
Bundle block storage fixup

imreallybadatnames:
Storage operations now manipulate internal storage state directly. Bundle contents and storage data are now synced with each other whenever needed.

Tested using a simple hopper setup (item insertion, item extraction, block break mid-transfer). The sum amount of extracted items and items within the bundle match the amount of inserted items.


DaFuqs (approved):
Works for me!

PR #451, opened 2024-08-11
Don't pass null to LootContextParameters.TOOL

f-raZ0R:
this causes Bad Things


DaFuqs:
this causes Bad Things

DaFuqs (approved):

PR #458, opened 2024-08-16
I'm swim

imreallybadatnames:
swim swom


Noaaan:
Looks good! Question remains whether we want to keep all the Spectrum fluids swimmable. I think it would make sense if you couldn't do so with Mud. Not sure about the rest

imreallybadatnames:
I feel that question goes to @DaFuqs (the one that created and populated this tag)

Noaaan:
Should be fine to merge for now. We'll change this based on feedback from playtesting

PR #459, opened 2024-08-17
You know how ghasts don't spawn ever

KatTheFox:
Adds a new head fusion recipe to create ghast heads because holy fuck finding them sucks. Also changes their head→memory recipe because why do they require one of their own drops that doesn't even always drop to be made into a memory that defeats the entire purpose.

### DIFF
--- a/src/main/resources/data/spectrum/modonomicon/books/guidebook/entries/creating_life/creature_manipulation.json
+++ b/src/main/resources/data/spectrum/modonomicon/books/guidebook/entries/creating_life/creature_manipulation.json
@@ -52,6 +52,10 @@
     },
     {
       "type": "spectrum:spirit_instiller_crafting",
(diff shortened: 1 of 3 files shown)


DaFuqs:
that's actually a cool idea for fusion

f-raZ0R:
I think this should involve nightdew in some way. You unsleep the phantom and it just becomes a regular ghost.

KatTheFox:
Nightdew isn't unsleeping though

f-raZ0R:
You sleep it even more then or something

f-raZ0R:
phantom has ties to insomnia
add sleep to phantom and it no longer has ties to insomnia and is sad now (this is why it cries now this is clearly the peak lore)

DaFuqs (approved):

PR #461, opened 2024-08-20
Add crushing recipes for turning clovers into Green dye

f-raZ0R:
I would _like_ for doing this with a regular clover to only have a percentage chance of giving you dye, due to the abundance of clovers, but sadly this recipe type does not support chanced outputs. 
Crushing a four leaf clover gives a lot more dye, and additionally a decent amount of XP.


DaFuqs:
One dye per item is more than ok I feel. Especially since Cactus is so easy to farm, once you find it.
Thanks!

PR #467, opened 2024-08-25
Revamp Deeper Down City Loot Tables

f-raZ0R:
Draft PR because I expect some tweaks to still be made as others give feedback on this.


DaFuqs:
Lots of great changes! Love the detail.
Seed vaults have a lot of loot chests. Might be good to tone down the amount of loot in each single one, so all combined it ends up as a manageable amount a player could carry in a single trip

f-raZ0R:
> Lots of great changes! Love the detail. Seed vaults have a lot of loot chests. Might be good to tone down the amount of loot in each single one, so all combined it ends up as a manageable amount a player could carry in a single trip

did that

PR #469, opened 2024-08-25
A few fixes for Sinytra Connector compat and broken Advancements.

eehunter:
Fixes mixin conflicts with Sinytra Connector, and refactors multiblock identifiers in advancement JSON to match Java code. 

Adds a try-catch to prevent crashing when running InkColors#elementals(). This is a temporary fix, and either this pull request will be updated or another will be created to implement a more proper solution.


DaFuqs:
Thank you lots!

DaFuqs (approved):

PR #475, opened 2024-09-01
Add mod compat recipe for Supplementaries Stasis Books

f-raZ0R:
Soap's been awfully mean to us, so we should appease it by implementing compat for the bubble blower enchant.
This will clearly make soap not cause more issues with turning things white, right?

I was a bit unsure whether to make this require blue ink and gems, or light gray and fragments like the vanilla treasure enchants. I've settled on it being the same structure as frost walker/mending, but i'm open to changing this. The stratine and palt balance each other out, creating a similar "stasis" effect as hoverblocks


f-raZ0R:
Github moment why did you close this pr when i renamed the branch

PR #478, opened 2024-09-05
Add more ways to get given the guidebook from interacting with spectrum

f-raZ0R:
Encourages more players to do spectrum stuff as soon as they find something from it


DaFuqs:
Not sure about the clover, since it does not have any magical properties, but the others are great.
Maybe even dragonbone, now that it is available from the start

f-raZ0R:
Done

PR #480, opened 2024-09-07
Make moonstone and its blocks whiter

Robotgiggle:
Moonstone exists to be the W part of the CMYKW spectrum. However, the updated moonstone textures coming in the 1.8 update seem much more blue than white. This changes them to use a whiter palette, closer to the old design, while still using the new textures.


Azzyypaaras:
image
bro.

Robotgiggle:
There is actually a difference if you look closely! Although tbf the ores are mostly just for completion's sake, since it would be kinda weird to re-palette everything else but leave them untouched

DaFuqs:
I take the 'bro' is not an objection?
The change is subtle for the most part, but I like it

Azzyypaaras:
I mainly found it funny that the moonstone ore got whiteified in spite of not having been touched at all in the resprite. The whole thing is suitably fine though I'd tweak the crystal and dust a bit.

While the criticism about it dipping a bit too much into the blues is fair, I do feel those two really benefit from having a vibrant blue rim light to accentuate the rest of the color. All of the gems (bar onyx kind of) have a splash of intense color to make the thing pop. However, since white has... you know, no saturation it needs something else to make the pure white pop.

Devoid of it, the shard and dust look kind of dead. That contrast does a lot of them, and I am of the opinion that the shards can be allowed to not stick 100% to their assigned color - they are after all minerals with flaws, and the role of pure color is left to the pigments.

DaFuqs:
Fair enough. I will leave this PR to you then

PR #482, opened 2024-09-09
Hideous reflection for QSL compatibility

unilock:
This PR commits crimes against the JVM to implement compatibility with Quilt Standard Library's `StatusEffectEvents`.  
This prevents a crash when QSL's `LivingEntityMixin` is applied before Spectrum's `LivingEntityPreventStatusClearMixin`, and fixes the latter not actually doing anything when the inverse is true.

We can't simply `modCompileOnly` QSL, since it refuses to remap under Fabric Loom. I guess it needs Quilt Mappings specifically? Whatever, this won't matter outside of 1.20.1 anyway.

TODO:
- [ ] Testing.


DaFuqs:
Implemented in #509

PR #483, opened 2024-09-09
Datafixers for blocks / items implementing InkStorage

unilock:
TODO:
- [x] Figure out how to register these things...
- [ ] Get InkStorageBlockFix working (???)
- [ ] Include more inventories as data fixing targets:
  - [ ] Spectrum's BlockEntities with Inventories
  - [ ] PlayerEntity Inventories
  - [ ] Arbitrary modded blocks (!?)

### DIFF
--- a/src/main/java/de/dafuqs/spectrum/SpectrumCommon.java
+++ b/src/main/java/de/dafuqs/spectrum/SpectrumCommon.java
@@ -7,6 +7,7 @@ import de.dafuqs.spectrum.compat.*;
 import de.dafuqs.spectrum.compat.reverb.*;
 import de.dafuqs.spectrum.config.*;
 import de.dafuqs.spectrum.data_loaders.*;
+import de.dafuqs.spectrum.datafixer.*;
 import de.dafuqs.spectrum.entity.*;
 import de.dafuqs.spectrum.events.*;
 import de.dafuqs.spectrum.explosion.*;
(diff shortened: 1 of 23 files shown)


unilock:
Since data fixers for items seem to be nigh-impossible to thoroughly implement and apply in an environment with other mods, perhaps a custom recipe could be used to convert the NBT data of pre-1.8 InkStorageItems?

A data fixer for InkStorageBlocks does appear to be necessary, however.

unilock:
Would it be feasible to implement InkStorage conversion during NBT deserialization, at least for InkStorageBlocks? As in, detect if the InkStorage is using old Ink IDs when `fromNbt` is called, do the conversion then, and maybe add a new NBT field that indicates that the block has been "datafixed".

Noaaan:
Could you explain the scope of this PR? Mainly wondering if this is targetting 1.7.0 -> 1.8.0 data

unilock:
@Noaaan
> Mainly wondering if this is targetting 1.7.0 -> 1.8.0 data

That is what the original goal of this PR was, yes.

DaFuqs:
> Would it be feasible to implement InkStorage conversion during NBT deserialization, at least for InkStorageBlocks? As in, detect if the InkStorage is using old Ink IDs when fromNbt is called, do the conversion then, and maybe add a new NBT field that indicates that the block has been "datafixed".

Certainly! Though that feels like a lot of work for something that would only be useful once.

unilock:
Closing in favor of #619 - datafixers are evil

PR #486, opened 2024-09-10
Add c: tool tags to spectrum items

f-raZ0R:
For SOME reason, there exist #c:axes , which is entirely separate from #minecraft:axes . Some mods are actually using this.
This PR adds these tags to spectrum's tools.


DaFuqs:
The `c:<tool>` tags are deprecated and were used before the minecraft ones were introduced in 1.19.4.
The other mods are outdated and should be fixed, instead.

PR #487, opened 2024-09-10
Add a couple miscellaneous dragonbone recipes

f-raZ0R:
dragonbone chunk -> bone meal, and bone ash -> ash


Azzyypaaras:
You know, bone ash to normal ash feels kind of unnecessary even conceptually. Bone chunks to bone meal is suitably funny though.

f-raZ0R:
Coming back to this pr, I suppose that makes sense due to the other ash recipe? But even still... both _are_ ash, and it makes sense that dragonrot would further melt the dragonbone into unrecognizability

DaFuqs:
- Added a missing file extension
- TODO: These recipes are not reflected in the guidebook yet

PR #500, opened 2024-10-01
Cosmetic blocks and bug fixes

GDPulsar:
First PR with some general additions and bug fixes.

Changes:

- Calcite and basalt crests will actually rotate correctly, and I checked other blocks and saw the pylon blocks also don't have the rotate or mirror functions, so I added it there too.
- Added the planed calcite and basalt slabs, stairs and walls. Stonecutter and crafting table recipes were added.
- Fixed the Dreamflayer's death messages having the wrong ID.
- I didn't realise that pizzer had already done the oversized item renderer fix, but I thought it was worth adding the commit anyway. I also changed the predicate tag name from "in_world" to "render_large" because it made more sense.


GDPulsar:
also i fixed the pyrite paneling having an id of `pyrite_panneling`. i think i got almost every usage but i'm not sure.

src/main/java/de/dafuqs/spectrum/registries/SpectrumBlocks.java
```diff
@@ -2259,13 +2286,28 @@ private static void registerStoneBlocks(FabricItemSettings settings) {
...
+		registerBlockWithItem("pyrite_plating_stairs", PYRITE_PLATING_STAIRS, settings, DyeColor.BROWN);
+		registerBlockWithItem("pyrite_plating_wall", PYRITE_PLATING_WALL, settings, DyeColor.BROWN);
 		registerBlockWithItem("pyrite_tubing", PYRITE_TUBING, settings, DyeColor.BROWN);
 		registerBlockWithItem("pyrite_relief", PYRITE_RELIEF, settings, DyeColor.BROWN);
 		registerBlockWithItem("pyrite_stack", PYRITE_STACK, settings, DyeColor.BROWN);
-		registerBlockWithItem("pyrite_panneling", PYRITE_PANNELING, settings, DyeColor.BROWN);
+		registerBlockWithItem("pyrite_paneling", PYRITE_PANELING, settings, DyeColor.BROWN);
+		registerBlockWithItem("pyrite_paneling_slab", PYRITE_PANELING_SLAB, settings, DyeColor.BROWN);
+		registerBlockWithItem("pyrite_paneling_stairs", PYRITE_PANELING_STAIRS, settings, DyeColor.BROWN);
+		registerBlockWithItem("pyrite_paneling_wall", PYRITE_PANELING_WALL, settings, DyeColor.BROWN);
```
DaFuqs: Renaming the 'panneling' typo now would mean all those blocks in the world will vanish, since they have a new block ID.
This is a change I noted for the 1.21 port, since we should not mess with players worlds, if not strictly necessary.

Hence why I fixed the lang already, but kept the block id as-is.

src/main/java/de/dafuqs/spectrum/registries/SpectrumBlocks.java
```diff
@@ -2259,13 +2286,28 @@ private static void registerStoneBlocks(FabricItemSettings settings) {
...
 		registerBlockWithItem("pyrite_plating", PYRITE_PLATING, settings, DyeColor.BROWN);
+		registerBlockWithItem("pyrite_plating_slab", PYRITE_PLATING_SLAB, settings, DyeColor.BROWN);
+		registerBlockWithItem("pyrite_plating_stairs", PYRITE_PLATING_STAIRS, settings, DyeColor.BROWN);
+		registerBlockWithItem("pyrite_plating_wall", PYRITE_PLATING_WALL, settings, DyeColor.BROWN);
 		registerBlockWithItem("pyrite_tubing", PYRITE_TUBING, settings, DyeColor.BROWN);
 		registerBlockWithItem("pyrite_relief", PYRITE_RELIEF, settings, DyeColor.BROWN);
 		registerBlockWithItem("pyrite_stack", PYRITE_STACK, settings, DyeColor.BROWN);
-		registerBlockWithItem("pyrite_panneling", PYRITE_PANNELING, settings, DyeColor.BROWN);
+		registerBlockWithItem("pyrite_paneling", PYRITE_PANELING, settings, DyeColor.BROWN);
+		registerBlockWithItem("pyrite_paneling_slab", PYRITE_PANELING_SLAB, settings, DyeColor.BROWN);
```
DaFuqs: Do we really need variants for everything?
Especially stairs and walls take up a LOT of block states

src/main/java/de/dafuqs/spectrum/registries/client/SpectrumModelPredicateProviders.java
```diff
@@ -247,12 +248,7 @@ private static void registerSlotReservingItem(Item item) {
 	}
 	
 	private static void registerOversizedItemPredicate(Item item) {
-		ModelPredicateProviderRegistry.register(item, new Identifier("in_world"), (itemStack, world, livingEntity, i) -> {
-			if (world == null && livingEntity == null && i == 0) { // REIs 'fast batch' render mode. Without mixin' into REI there is no better way to catch this, I am afraid
-				return 0.0F;
-			}
-			return currentItemRenderMode == ModelTransformationMode.GUI || currentItemRenderMode == ModelTransformationMode.GROUND || currentItemRenderMode == ModelTransformationMode.FIXED ? 0.0F : 1.0F;
-		});
+		ModelPredicateProviderRegistry.register(item, new Identifier("oversized"), (itemStack, world, livingEntity, seed) -> seed == 817210941 ? 1.0F : 0.0F);
```
DaFuqs, line 251: 'oversized' does not really reflect what the predicate does.
It does not specify that the item should be rendered oversized at all, it specifies if a model is rendered in a gui or in-world.

DaFuqs (changes requested):
- Added a few notes
- The fixes and changes to planed calcite/basalt are nice
- The new items are not added to the guidebook yet

src/main/java/de/dafuqs/spectrum/registries/SpectrumBlocks.java
GDPulsar: i mean, i was only thinking of doing planed calcite and basalt and then azzyy challenged me to do pyrite as well in insider trading. i can remove this if need be

src/main/java/de/dafuqs/spectrum/registries/client/SpectrumModelPredicateProviders.java
GDPulsar: image

Noaaan:
To verify: does this PR add any new paneling blocks? If this is the case, these should be using the correct ID to begin with so we don't literally create more work for ourselves later. Only the existing paneling blocks need to remain unchanged

Azzyypaaras:
Pulsar, would you be so kind as to solve the pyrite tag conflict?

src/main/java/de/dafuqs/spectrum/registries/client/SpectrumModelPredicateProviders.java
Noaaan: I think it is fine for now. I agree that it should be renamed, but we can revisit it later. I use this snippet in other mods - maybe extracting that API to a small library is in order?

f-raZ0R:
Any updates on this PR?
Seems that it is just missing guidebook entries, to my understanding?

GDPulsar:
it is no longer missing guidebook entries

PR #502, opened 2024-10-03
Hide Manxi from recipe viewers

f-raZ0R:
she should be a Surprise :3


DaFuqs:
Why does it exist in item form in the first place? o.O

PR #503, opened 2024-10-03
Make Falling Colorful Shooting Stars not implode your game if a mod i…

f-raZ0R:
We have this helper for the specific purpose of not imploding your game if a mod extends the dye enum


DaFuqs (approved):
Good catch!

PR #504, opened 2024-10-04
Move Noxwood Recipes to the crafting table

f-raZ0R:
Fixes the inconsistency of basic wood stuff being only makeable on the pedestal, ONLY for noxwood.


DaFuqs:
TODO: The changes still have to be reflected in the guidebook.
I will take care of it.

PR #507, opened 2024-10-05
Ender that actually Radiates

Robotgiggle:
The new texture for Radiating Ender is very cool, but it doesn't really convey the idea of "radiating" at all. As such, I decided to make a new one that does, based on the old nether reactor core from Pocket Edition. This version is also animated and pulsates slightly.

image


DaFuqs:
@Azzyypaaras this one is for you

PR #509, opened 2024-10-06
Hideous ASM for QSL compatibility

imreallybadatnames:
A so-called "successor" to PR #482 that does the exact same thing as before but using a different flavor of war crimes.
All that this ASM wizardry does is emit code that does not rely on reflection in the slightest bit, thereby moving certain runtime guarantees into JVM verify-time. The decompiled version of the Mixin'd class looks as if the QSL event register was added using ordinary means, without any Mixins.
~~TODO: check if it doesn't explode violently~~ done


unilock:
I am horrified. Well done!

PR #511, opened 2024-10-06
Fix Amaranth-FD Compat but actually

f-raZ0R:
Sadly because this loads before integration packs it can't require it to be enabled


f-raZ0R:
It now requires the integration pack to be enabled 🥳

DaFuqs:
Made a few small changes.
Feel free to test. I tested with and without Farmers Delight and it seemed to work flawlessly

PR #515, opened 2024-10-09
Cloak Mermaids Gems as Kelp

f-raZ0R:
why are they even cloaked as orange dye??


DaFuqs:
~~because they are orange~~
You know, that is a good point

PR #516, opened 2024-10-10
Mermaid Gems can be Extracted for water in many mods + Fix Titration Barrel recipes with milk/honey breaking

f-raZ0R:
this pr does both of those things

~~KNOWN ISSUES (that aren't on our end so I can't fix):
This current implementation causes the game to crash if a mermaid gem is thrown onto a create item drain.
This is an error on create fabric's end, and there is nothing I can do about this.~~
fixed on create's end


f-raZ0R:
... are the builds failing because of matchbooks AGAIN

src/main/java/de/dafuqs/spectrum/blocks/fluid/RemainderlessItemFluidStorage.java
```diff
@@ -0,0 +1,82 @@
...
+        this.containedAmount = containedAmount;
+    }
+
+    @Override
+    public long extract(FluidVariant resource, long maxAmount, TransactionContext transaction) {
+        // If the context's item is not fullItem anymore, can't extract!
+        if (!context.getItemVariant().isOf(fullItem)) return 0;
+
+        // Make sure that the fluid and the amount match.
+        if (resource.equals(containedFluid) && maxAmount >= containedAmount) {
```
DaFuqs, line 37: Does that mean that if a block wants to extract less than a containers worth - in this case a bucket - at a single time, extraction always fails?

Kind of a big issue, if another mod only tries to extact 100 droplets each tick.

imreallybadatnames: This isn't really solvable as it is now. It can only be worked around by e.g. extracting the fluid first, and then using a secondary storage containing the resultant fluid for further processing.

Noaaan: For now this should be fine as it only happens with Mermaids Gems. This type of behavior would apply when using say, a Lava Bucket on a partially filled tank or cauldron.

DaFuqs:
Uuuuh, with the fluid stuff fixed why should we have that `titration_never_fluid` tag now and all the fluid recipes removed ಠ_ಠ

f-raZ0R:
I made this PR before that was fixed, so we can add those back. The tag is still needed because there are some recipes that use mermaid's gems in the titration barrel, unless we want to change mint schnapps and the other ones to not use gems

PR #524, opened 2024-10-21
Part of manor loot

f-raZ0R:
Adds loot for six chests in the undergrowth manor.
Also fixes up bumblezone compat because some tags didn't have replace:false in them


src/main/resources/data/spectrum/loot_tables/chests/undergrowth_manor/basement_resting_grounds.json
Noaaan, line 1: I see this is empty - Is it unfinished?

src/main/resources/data/spectrum/loot_tables/chests/undergrowth_manor/basement_heavens_cradle.json
```diff
@@ -0,0 +1,142 @@
...
+      ]
+    },
+    {
+      "rolls": 2,
+      "entries": [
+        {
+          "type": "minecraft:item",
+          "name": "minecraft:dragon_egg",
+          "weight": 0,
+          "quality": 3
```
Noaaan, line 98: Are we sure about this one? Could potentially cause issues in a lot of servers where the Dragon Egg is coveted

src/main/resources/data/spectrum/loot_tables/chests/undergrowth_manor/basement_sinners_outcrop.json
```diff
@@ -0,0 +1,141 @@
...
+      ]
+    },
+    {
+      "rolls": 1,
+      "entries": [
+        {
+          "type": "minecraft:item",
+          "name": "spectrum:nectardew_burgeon",
+          "weight": 666,
+          "quality": 666
```
Noaaan: While humorous, I want to note that setting these values will cause the burgeons to be affected by luck. A single point of negative luck (maybe via Bad Luck) will cause this to be unfindable. 
Otherwise it should function as expected, unless something something mod compat shenanigans.

src/main/resources/data/spectrum/loot_tables/chests/undergrowth_manor/basement_heavens_cradle.json
```diff
@@ -0,0 +1,142 @@
...
+      ]
+    },
+    {
+      "rolls": 1,
+      "entries": [
+        {
+          "type": "minecraft:item",
+          "name": "spectrum:aether_vestiges",
+          "weight": 413,
+          "quality": 612
```
Noaaan: See the similar comment below regarding quality and luck

Noaaan (commented):
I will assume the NBT for the structures is good. Nice work!
Added some comments regarding a few of the loot entries, as well as the empty loot table

src/main/resources/data/spectrum/loot_tables/chests/undergrowth_manor/basement_resting_grounds.json
f-raZ0R: Indeed it is

src/main/resources/data/spectrum/loot_tables/chests/undergrowth_manor/basement_sinners_outcrop.json
f-raZ0R: Oh this is true, i just set these to arbitrarily high numbers to try to make sure that they don’t get overwritten by some other mod injecting into the loot table, combined with high luck
(looking at YOU simply swords)
the concern with unluck is… valid though.

src/main/resources/data/spectrum/loot_tables/chests/undergrowth_manor/basement_heavens_cradle.json
f-raZ0R: Azzy specifically requested this one, bring it up with her

PR #525, opened 2024-10-22
Changed Pastel Network & Black Hole Chest filters to use ItemVariant

GDPulsar:
Just a small change this time


Noaaan (approved):
Looks good overall. Could you quickly summarize why we wanted to make this change? Have been a bit out of the loop regarding it 😅

Azzyypaaras:
> Looks good overall. Could you quickly summarize why we wanted to make this change? Have been a bit out of the loop regarding it 😅

Oh, I can explain.
When I was working on the Pastel overhaul and tag filtering, that required some changes to how filters are handled. Originally they were just items, but I needed nbt access so I swapped them to work off of ItemStacks (amidst a few other minor things).

Daf later pointed out that since item count is not being used, it makes more sense to use ItemVariants over ItemStacks as the former does not store count whereas the later does.

PR #532, opened 2024-10-27
Replace Overabundance of Eggs in Pedestal Ruins with suspicious gravel

f-raZ0R:
Now every ruin has a max of 1 (one) egg


DaFuqs:
Yes, this is nice. I'll have to look at the structures ingame since github does not display them, but I am sure those are fine.

DaFuqs:
1. Should the loot table `spectrum:archaeology/pedestal_ruins_stone_gravel` you set actually exist? Since it does not for me
2. The suspicious gravel in the stone variant is set 2 blocks below the pedestal. Intended? (in red here)
image

f-raZ0R:
> 1. Should the loot table `spectrum:archaeology/pedestal_ruins_stone_gravel` you set actually exist? Since it does not for me
> 
>     2. The suspicious gravel in the stone variant is set 2 blocks below the pedestal. Intended? (in red here)
>        image

,, wait hang on what.
why didn't those get put into the commit

PR #534, opened 2024-10-27
Decay away makes blackslag in the DD instead of Dirt

f-raZ0R:
Fixes the lag problem of the dirt decay away makes getting turned back into black materia by the midnight solution, and then reconverted by the decay away, in an endless cycle


f-raZ0R:
I'm thinking about this more, and actually, hm
I think it'd be better to refactor this to be data-driven?
Maybe even with the ability for decay away to make different blocks depending on current height (e.g in the overworld, dirt above y 50, stone above y0, and deepslate otherwise)?

but for now at least, this will help with the black langasts

DaFuqs:
Yeah, I had the same in mind with making it data driven some time, but for now it's absolutely fine

PR #538, opened 2024-10-30
Fix sleep brew unlocks + various guidebook improvements

Robotgiggle:
Unlock stuff:
- All sleep brews except Evernectar are now locked behind age confirmation since they're all alcoholic
- Evernectar is not locked behind age confirmation since it's a crafting ingredient and clearly isn't meant to be consumed
- Verdigris wine is also locked behind quitoxic reeds since they're used in the recipe
- Removed unused extra `unlock_glass_peach_titration` advancement
- Immunity recipe unlocks from the internal `hidden/collect_cookbooks/poisoners_handbook` advancement like all the other poisoner book recipes, instead of the user-facing `lategame/collect_poisoners_handbook` advancement

Guidebook stuff:
- Immunity recipe appears in the Brewing Potions section once unlocked, with a note linking back to the poisoner book
- Lethal slumber page in the Brewing Potions section explains that the recipe only works for potion fillables
- Nature's Staff guidebook entry no longer mentions brown, since it's not used in the recipe anymore
- Verdigris Wine entry mentions that it lets you sleep through the day, since it's not otherwise clear why you'd want to make it


DaFuqs (approved):
Ty lots!

PR #539, opened 2024-10-31
Droopleaf and some other stuff

f-raZ0R:
Draft PR so I can A: Resolve merge conflicts, and B: better art/guidebook entries haven't been made yet (i'll get to the guidebook soon)

also does a couple other things that i got a bit distracted with, like letting Lily Pads be placed on Mud or Liquid Crystal (and adding a fluid dipping recipe to cut off the stem of a big dripleaf to be left with a lilypad), or making more blocks Flammable.

Adds the Droopleaf, certified Doohickey. I've posted a bunch of videos of this in dev channels/insider-trading, you already know what this does.


src/main/java/de/dafuqs/spectrum/mixin/FoxEntityMixin.java
```diff
@@ -11,7 +11,7 @@
 @Mixin(FoxEntity.class)
 public class FoxEntityMixin {
 
-    @ModifyReturnValue(method = "isSleeping", at = @At("RETURN"))
+    @ModifyReturnValue(method = "isSleeping()Z", at = @At("RETURN"))
```
imreallybadatnames, line 14: nit: but why though? It works as it currently is just fine, no issues here. Unless there is? Is it related to mod compat?

f-raZ0R: It wasn't working correctly, so I tried adding this and I guess I never undid the change.

src/main/resources/data/spectrum/worldgen/configured_feature/bristle_sprouts.json
```diff
@@ -28,11 +28,6 @@
           "type": "minecraft:block_predicate_filter",
           "predicate": {
             "type": "minecraft:matching_blocks",
-            "offset": [
-              0,
-              -1,
-              0
-            ],
             "blocks": "spectrum:sawblade_grass"
```
DaFuqs: Wait, wouldn't the predicate search for air and sawblade grass at the same pos now? o.O

f-raZ0R: It still generates successfully, due to mojank? But yeah that is a good point, the sawblade grass check might not be doing anything here to begin with, seeing how it still generates despite being seemingly impossible if this was doing something

f-raZ0R: Wait never mind I must've been very tired when i was testing it, yeah this just makes it never generate at all, I was confusing it with sawblade holly bushes, yeah this needs to be fixed

DaFuqs:
Looks like Github set the PR to merged, since all commits of the PR were added to main, despite some of them being reverted.
For the Droopleaf we'll need to create another PR later.

PR #540, opened 2024-10-31
Fix other new alcohol unlocks + couple extra things

Robotgiggle:
Oops, missed these three in the last PR. All new alcohols should now be properly age-locked.


DaFuqs (approved):
<3
Looks perfect, ty!

PR #542, opened 2024-11-02
Various fixes for kindling guidebook entries

Robotgiggle:
- Kindling art page is now hidden until you've hatched a kindling
- Hint about sleep effects and kindlings is now hidden until you've both hatched a kindling and obtained somnolence
- Hint about obtaining feathers with shears is back (this seems to have been accidentally removed when the art page was added)
- Effulgent Feather entry is now hidden until you obtain a feather, so it doesn't appear early with a shrouded item


DaFuqs:
Thank you!

PR #545, opened 2024-11-03
Greatly simplify Azurite Aura particle rendering logic

imreallybadatnames:
In total, gets rid of 6 unnecessary object allocations. Comparing this to the original quaternion rotation math, this is, at least, a ~50x reduction of math operations needed. Renamed the variables to be a little more readable, and made length final for good measure.


DaFuqs:
Looks absolutely identical with a good improvement in performance. Ty!

PR #546, opened 2024-11-03
Remove Recipe and Book entry for the Primordial Lighter

f-raZ0R:
Why does a creative-exclusive item not intended for survival use, have an unobtainable recipe and _book entry_?


DaFuqs:
It was once obtainable and since the entries were already there, I figured we could keep them in case a pack wants to add an unlock for it.
Opinions?

f-raZ0R:
If a pack wants to add an unlock for it, couldn't they just datapack in their own recipe and book entries?

DaFuqs:
Absolutely. But as we all know they would probably be sub-par and the guidebook entry nonexistent :P

f-raZ0R:
Why don't we add recipes and guidebook entries for all the other creative only items too then? :P

f-raZ0R:
But anyways, as for the actual point, there is zero reason to have these recipes or guidebook entry in the game, as they are unavailable without commands, and only serve to confuse players about why they still have a recipe involving gloves they haven't unlocked yet, just for the niche scenario of a pack implementing an unlock for the item, yet not bothering to implement their own recipe or guidebook entry? In which case, that sounds like a lazily designed pack and we shouldn't have this just to provide to them

DaFuqs:
Having a recipe that you could never unlock is a valid argument, you are right

PR #548, opened 2024-11-04
Resonance drops don't stop searching for entries after finding one

f-raZ0R:
Fixes issue where having create compat enabled meant the pure zinc drop overrode all the other pure resource drops


DaFuqs (approved):

DaFuqs:
Looking at the code for the ModifyDrops & DropSelf Processors, there can still be lots of issues if two or more conditions match. But for now we do not have any of those combinations

PR #550, opened 2024-11-05
Add more furnace fuels

f-raZ0R:
Added fuel times:
Vegetal: Smelts 4 items
Vegetal Block: Smelts 40 items (gives a slight bonus, like coal blocks)

Pure Alcohol/Chrysocolla: Smelts 250 Items. Don't try this at home

Orange Pigment Specifically (as it represents heat): Smelts 12 items

Incandescent Essence: Smelts 12 items, dubious on whether we should add this, as azzy makes a decent argument for not having this usable as fuel

Side note: due to being in the #non_flammable_woods tag, Weeping Gala wood can not be used as a furnace fuel. We should really come to a consensus on whether weeping gala should be flammable or not, as this is currently quite inconsistent in the code base

### DIFF
--- a/src/main/java/de/dafuqs/spectrum/registries/SpectrumItems.java
+++ b/src/main/java/de/dafuqs/spectrum/registries/SpectrumItems.java
@@ -995,15 +995,14 @@ public class SpectrumItems {
 
 		FuelRegistry.INSTANCE.add(SpectrumItems.PURE_COAL, 3200);
 		FuelRegistry.INSTANCE.add(SpectrumBlocks.PURE_COAL_BLOCK, 32000);
-
+		
 		FuelRegistry.INSTANCE.add(SpectrumItems.VEGETAL, 800);
 		FuelRegistry.INSTANCE.add(SpectrumBlocks.VEGETAL_BLOCK, 8000);
-
-		FuelRegistry.INSTANCE.add(SpectrumItems.PURE_ALCOHOL,50000);
-		FuelRegistry.INSTANCE.add(SpectrumItems.CHRYSOCOLLA,50000);
+		
+		FuelRegistry.INSTANCE.add(SpectrumItems.PURE_ALCOHOL, 16000);
+		FuelRegistry.INSTANCE.add(SpectrumItems.CHRYSOCOLLA, 16000);
 
 		FuelRegistry.INSTANCE.add(SpectrumItems.INCANDESCENT_ESSENCE, 2400);
-		FuelRegistry.INSTANCE.add(SpectrumItems.ORANGE_PIGMENT, 2400);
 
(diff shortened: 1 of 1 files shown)


Robotgiggle:
I do like that orange pigment works as a fuel, but I think it's a little too powerful for something you can grind out en masse. I'd suggest making it equal to coal at the very best.

Also, while smelting tons of items with pure alc or chrysocolla is very funny, I think 250 is just way too much. That's nearly four stacks of items - when was the last time you cooked that much stuff in a single furnace? I feel like much of that massive burn time would end up getting wasted. I'd suggest lowering that one to either 64 or 100.

DaFuqs:
- Vegetal: I'm imagining it kinda mushy. Not sure it should burn?
- High % alcs: Having them burn sounds kinda fun! Though I agree a value for 64 items would probably match better, being a bit lower than a bucket of literal lava
- Incandescent Essence: I could see it being used for heating. Though if we do that, Frostbite essence should also have a use
- Same for Pigment. Only Orange Pigment having a secondary use feels weird. Maybe we can find a minor use for each one? Lime could maybe be used as Bone Meal, while Magenta ones trigger a random tick

Robotgiggle:
Using lime pigment as bonemeal would make the Nature's Staff obsolete, since that's pretty much what it already does

Robotgiggle:
Maybe a better solution for the orange pigment would be to make some kind of ink fillable that you charge with orange ink and then put in a furnace to act as fuel, sort of like the Lava Crystal from Blood Magic

f-raZ0R:
I would like to point out that we are in "feature freeze" and expecting to release a version later this week

DaFuqs:
> Maybe a better solution for the orange pigment would be to make some kind of ink fillable that you charge with orange ink and then put in a furnace to act as fuel, sort of like the Lava Crystal from Blood Magic

That's a very cool idea, ngl. I will note that down for later, though using Ink instead.

Robotgiggle:
> I would like to point out that we are in "feature freeze" and expecting to release a version later this week

Yeah, this particular idea would definitely not be included in 1.8. I was just proposing it as a way to keep some kind of furnace-related use for Orange while removing the direct fuel value.

f-raZ0R:
Re: nerfing pure alch burn time:
I had it so high mainly due to the comparative difficulty in obtaining it. It is very easy to get a lot of lava quickly, it is very hard to get a lot of pure alcohol quickly. If pure alcohol is worse than lava... why use it when you could just use lava instead?

Robotgiggle:
That's a fair point, but tbh I don't really see anyone _ever_ making pure alc specifically to use as furnace fuel, even with the super high burn time. The cost of making it is so high in both time and materials that I feel like sticking it in a furnace would only ever be done for the memes - you could use all the time it takes to make pure alc to obtain more than enough of any other fuel to balance out even the 250 item burn time.

PR #552, opened 2024-11-08
Mod Compat Guidebook Entries

f-raZ0R:
… spelunkery compat minorly, + Supplementaries Ash interop

still haven't added malum stuff/pure zinc/bumblezone enchantabilities+ enchant recipes yet


f-raZ0R:
Individual mod blurbs kinda suck but it's hard to come up with things

PR #554, opened 2024-11-08
Allow eternal slumber fillables

Robotgiggle:
Making potion fillables was disabled for Eternal Slumber to prevent Nightfall's Blade cheese, but the Nightfall's Blade already has code to make strong sleep effects (eternal and fatal slumber) cost way more. As such, I think it's fine to enable fillable for Eternal Slumber. Also, this means you can now use it on Concealing Oils.


DaFuqs:
I approve more tomfoolery

PR #555, opened 2024-11-08
Revert "bump nightdew drop chances"

f-raZ0R:
we should not be making changes like this right before release without any input


Azzyypaaras:
Don't try to sidestep me.

PR #556, opened 2024-11-08
Fix Biome Makeover Compat page missing icon

f-raZ0R:
antidotes aren't their own item 🙈


Robotgiggle:
Might as well fix the overlapping icons while you're at it
https://discord.com/channels/906264467636109333/911958718067929118/1304572497391390791

f-raZ0R:
done

PR #559, opened 2024-11-09
Fix kindling advancements

Robotgiggle:
Now that we have the ability to turn memories into heads, it's possible to bypass the "hatch a kindling from a memory" check by turning your memory into a head and then a spawner, which prevents you from unlocking feathers or bloodstone until you hatch a kindling the "proper" way. This fixes that issue by allowing `spectrum:remember_kindling` to trigger when you interact with a kindling in-world, so long as you've already unlocked the memory recipe. It also makes `spectrum:pluck_effulgent_feather` hidden until you unlock it, both to prevent spoiling the **RUN** and to prevent it from appearing in the advancement tree with a shrouded item for an icon.

The change to `spectrum:remember_kindling` does mean that it will trigger immediately upon unlocking the memory recipe if you've already interacted with a kindling somehow (ie someone else's kindling on a multiplayer server), but I think that's an OK price to pay in exchange for fixing the spawner bypass.

### DIFF
--- a/src/main/resources/data/spectrum/advancements/hidden/entity_interact/kindling.json
+++ b/src/main/resources/data/spectrum/advancements/hidden/entity_interact/kindling.json
@@ -1,0 +1,50 @@
+{
+    "criteria": {
+      "entity_interact": {
+        "trigger": "minecraft:player_interacted_with_entity",
+        "conditions": {
+          "player": [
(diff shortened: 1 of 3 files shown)


DaFuqs:
Sounds about right, ty!

PR #564, opened 2024-11-10
Implement Droopleaf (for real this time)

f-raZ0R:
Second Verse, same as the first.


DaFuqs:
Converted to draft, since it still lacks textures

DaFuqs:
Closing that one for now, since it's not in a state to include and the branch it targets is pretty outdated now.

PR #570, opened 2024-11-12
Update MermaidsBrushBlock.java trying to fix multiple gems drop.

Ethanolgie:
This is a strange but simple bug,When normally getting random ticked,MermaidsBrush should spawn item when the age 'technically reach 8'(7+1),but it will spawn item when the age reachs 7,7 % 8 is still 7 but it will drop a gem.So sometimes it will drop gems if a bone meal made it age=7,and drop again when it grows again.


DaFuqs (approved):
true! thank you

PR #571, opened 2024-11-12
Fix infinite Meat exploit

f-raZ0R:
Rotten flesh can be crushed into multiple meat scraps. Which, using this recipe, can be turned back into rotten flesh...


DaFuqs:
How critical is infinite meat scrap for neepmeat? Do we *want* to fix this?

Infinite rotten flesh sounds like a "fun mod interaction" that does not hurt too much.

f-raZ0R:
It.. kinda does completely overstep a big part of neepmeat's gameplay loop by being able to get infinite meat so trivially

DaFuqs:
Oh. Well then, sure!

PR #572, opened 2024-11-13
Blacklist colored preservation stone from paintbrush recoloring

Robotgiggle:
Currently, colored preservation stone gets caught by the `getCursedBlockColorVariant` system, and thus it's a valid target to be recolored using a paintbrush. Given that preservation stone is an unbreakable structure component, I don't think it makes much sense for players to be able to just recolor it at will - particularly since it allows you to (intentionally or otherwise) change the color of the "backup" preservation stone above the roundels in mixing ruins. 

~~This PR adds colored preservation stone to the ink_effect_blacklisted tag so it can no longer be recolored.~~
This PR adds a hardness check to getCursedBlockVariant so that it always fails on unbreakable blocks.

### DIFF
--- a/src/main/java/de/dafuqs/spectrum/helpers/BlockVariantHelper.java
+++ b/src/main/java/de/dafuqs/spectrum/helpers/BlockVariantHelper.java
@@ -25,7 +25,7 @@ public class BlockVariantHelper {
 		
 		BlockState blockState = world.getBlockState(blockPos);
 		
-		if (blockState.isIn(SpectrumBlockTags.INK_EFFECT_BLACKLISTED)) {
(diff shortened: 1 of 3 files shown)


DaFuqs (approved):
Looks perfectly fine

DaFuqs:
Agreed, definitely.
I feel we should also hard-exclude all unbreakable blocks by default (checking the hardness in-code).
Opinions?

Noaaan (commented):
I can't think of any common use-cases of recoloring unbreakable blocks. It could likely break with other mods if they have similar puzzle components.

Robotgiggle:
Hardness check has been implemented. I also removed the colored pres stones from the `ink_effects_blacklisted` tag since that's redundant now.

PR #574, opened 2024-11-13
Various advancement fixes

### LINKED ISSUE #573: [BUG] Three advancement bugs
**Describe the bug**
Heavily Delighted triggers when you put only one of the gravity rings on.
Once Intertwined triggers prematurely when you enter the Deeper Down after obtaining a jade bulb.
An Artist’s Vanitas triggers without needing to craft any legendary weapon, only requiring their recipes to be unlocked, I think.

**To Reproduce**
I don’t know what to say.

**Expected behavior**
Heavily Delighted should’ve been obtained by equipping both kinds of gravity rings at the same time.
Once Intertwined should’ve only been obtained when crafting a jade bulb using the secret recipe.
An Artist’s Vanitas should’ve been triggered once you obtain all of the legendary weapons.

**Minecraft version**
1.20.1

**Mod version**
1.8.0

Robotgiggle:
Fixes #573

### DIFF
--- a/src/main/java/de/dafuqs/spectrum/blocks/spirit_instiller/SpiritInstillerBlockEntity.java
+++ b/src/main/java/de/dafuqs/spectrum/blocks/spirit_instiller/SpiritInstillerBlockEntity.java
@@ -202,20 +202,24 @@ public class SpiritInstillerBlockEntity extends InWorldInteractionBlockEntity im
 		}
 		
 		if (lastInteractedPlayer instanceof ServerPlayerEntity serverPlayerEntity) {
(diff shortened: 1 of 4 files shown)


DaFuqs:
Looks good, thank you!

The change to jade vine crossbreeding (and the pre existing hardcoded unlock for boss memories) makes me wonder if crafting advancements should have the option to trigger based on the recipes `group` property.
That way we could just trigger if a recipe in the group `boss_memories` or `jade_vine_crossbreeding` was crafted, without using hardcoded advancement ids.

PR #577, opened 2024-11-17
Add sarsaparilla, revamp fishing loot

f-raZ0R:
This PR:
-adds a new (Non-Alcoholic!) infused beverage, Sarsaparilla, made with clovers. To make up for the mint beverages being changed to both made with four leaf clovers.
-Overhauls fishing loot in fluids besides water! Previously, this loot was rather.. lackluster, resulting in not much reason to do so, besides getting crawfish/koi.
I'm finally giving people the renewable shimmerstone they want :P 
-Boosts rate of nightdew in trail ruins, by adding them to the trail_ruin_common loot table as well. Should make excavating trail ruins more rewarding earlygame


DaFuqs (approved):
juicy

PR #578, opened 2024-11-17
Add malachite smelting recipes

f-raZ0R:
Malachite ore... can't be processed at all. This very much sucks if your only tool you have of a high enough mining level has silk touch (example, the bedrock pickaxe which comes with silk).

This PR fixes that, adding the missing smelting recipes for the ore


DaFuqs:
Neat!

PR #586, opened 2024-11-19
Scarred prevents animals from growing up

f-raZ0R:


DaFuqs:
A way to make animals not grow up would be cool.

Not sure Scarred is the right way, though, since the animal continues to grow up once the effect runs out.

PR #587, opened 2024-11-19
Item Unlock Tweaks + Neat Ring Recipe change :tomfoolery:

f-raZ0R:
Fixes a lot of recipe unlocks where you unlock the recipe before having unlocked everything used to make it, and vice versa for cases involving pure resources, as pure resources can _also_ be obtained through resonance.
Also seperated out ampoule unlocks so as soon as you can make azurite ones, you unlock azurite ones, without needing to get _refined bloodstone_ just to be able to make azurite ampoules.

Oh, also changed the neat ring recipe to use pure gold instead of regular gold. Because we clearly need to upgrade the recipe every update :3


src/main/resources/data/spectrum/advancements/unlocks/items/glass_ampoules.json
```diff
@@ -1,34 +1,29 @@
...
+			"trigger": "revelationary:advancement_gotten",
+			"conditions": {
+				"advancement_identifier": "spectrum:midgame/create_refined_azurite"
+			}
+		},
+		"unlock_bloodstone_ampoule": {
+			"trigger": "revelationary:advancement_gotten",
+			"conditions": {
+				"advancement_identifier": "spectrum:grow_bloodstone_in_crystallarieum"
+			}
```
DaFuqs: those advancement checks should prob. refer to the amoule unlock advancements instead

DaFuqs (commented):
Lots of good stuff, thanks!
That is a huge bunch of fixes.

Also yes to the neat ring recipe change. It shall be tradition. 🙏 

A few things I am not completely sure about:
- Ampoule unlock change: Now unlocked when collecting any of the required resources for each recipe instead of all 3 of Azurite, Malachite, Bloodstone. I feel the unlock was a nice reward for collecting the whole set and immediately getting all 3 amoules to play around with, instead bit by bit. But having always at least the Azurite one when getting hummingstone is a fair point, too. Opinions?
- Knowledge Gem unlock change: Was changed from touching liquid crystal to building the Onyx Tier Pedestal. The Onyx Pedestal unlocks quite a bunch of stuff already, so the unlock was there right next to the Enchanter since they go hand in hand and the unlocks do not down in all the start-of-onyx-tier unlocks. Not sure I vibe with that change.

src/main/resources/data/spectrum/advancements/unlocks/items/glass_ampoules.json
f-raZ0R: .... how did i miss this

unilock:
This PR classloads `BotaniaItems` before Botania initializes, causing the game to crash during startup :(

f-raZ0R:
it What.
Oh god it didn't do this in a dev instance

unilock:
Quoth the Fabric Wiki:

> *[Entrypoints - A note about load order and phases (or a lack thereof)](https://wiki.fabricmc.net/documentation:entrypoint#a_note_about_load_order_and_phases_or_a_lack_thereof)*
> The order in which mod initializers are called is undefined, and cannot be influenced so that mod A's initializers are called after mod B's initializers, or so that mod C's initializers are called before mod D's initializers.

PR #588, opened 2024-11-19
Make Modular Explosives easier to craft, and add them to the Equipment Tab in the book as well

f-raZ0R:
As parametric mining devices are equipment too.

### DIFF
--- a/src/main/resources/data/spectrum/advancements/unlocks/blocks/modular_explosives.json
+++ b/src/main/resources/data/spectrum/advancements/unlocks/blocks/modular_explosives.json
@@ -1,9 +1,9 @@
 {
   "criteria": {
-    "collected_stratine_gem": {
+    "collected_stratine": {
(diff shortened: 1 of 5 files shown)


DaFuqs (approved):
- The change from Stratine Gems => Fragments is a very good one. That makes them much more easily accessible
- Maybe it makes more sense to have the guidebook entry in the equipment tab only, not magical blocks? If we have them in both we should make use of a `spectrum:entry_link` type entry ([Example](https://github.com/DaFuqs/Spectrum/blob/1.20.1-aria-for-painters/src/main/resources/data/spectrum/modonomicon/books/guidebook/entries/decoration/shards.json)), so we do not have to duplicate the whole entry

f-raZ0R:
.... i was not aware we had that entry type, that's a good idea

f-raZ0R:
Done

PR #591, opened 2024-11-20
Make 1.8 plants compostable

Robotgiggle:
Adds composter values for all the new plants in 1.8, along with Resonant Lillies and Blood Orchids which didn't have them for some reason.

Also reorganizes the `SpectrumCompostableBlocks` file to sort the entries by biome.


DaFuqs (approved):
Awesome. Thank you!

PR #594, opened 2024-11-21
Variety of potion and potion-related tweaks

Robotgiggle:
- Immunity recipe in guidebook now mentions low yield
- Vanilla fire resistance recipe in guidebook now directly states that all vanilla recipes work
- Removed potency nerf from absorption recipe
- Other recipes with potency nerf (resistance, wither, toughness, somnolence) now mention it in their guidebook page
- Compat recipes (and book entry) for the new hexcasting potions
- All drinks can now be consumed even while at full hunger (among other things, this makes restoration tea much more usable)
- When attempting to cure an incurable effect, the amount of duration removed is now either 1 minute or 60% of the current duration, whichever is smaller
- When fatal slumber is cured via any method other than a command, it turns into eternal slumber instead of just vanishing
- Immunity now works to cure and block fatal and eternal slumber
- When immunity cures or blocks fatal slumber, the immunity is removed too (leaving you with eternal slumber and no immunity)
- When immunity cures or blocks eternal slumber, it loses 30 seconds of duration, just like when it blocks an incurable effect
- Replaced the `immunity_immune` and `unclearable` tags with `bypasses_immunity` and `bypasses_whispy_circlet`, and disentangled their functionality so each one only applies to the relevant mechanic
- Added new tag `bypasses_nectar_gloves`, which does exactly what it sounds like 
- Nectar gloves and whispy circlet being unable to block/cure soporific effects is now handled by including the `soporific` tag in their respective bypass tags rather than being hardcoded
- Nectar gloves no longer waste dike on effect blocking if you have immunity (so long as the incoming effect isn't in `bypasses_immunity`)


src/main/java/de/dafuqs/spectrum/mixin/EffectCommandMixin.java
```diff
@@ -25,6 +26,9 @@ private static void clearIncurableEffects(ServerCommandSource source, Collection
             for (StatusEffectInstance effect : living.getStatusEffects()) {
                 if (((Incurable) effect).spectrum$isIncurable())
                     ((Incurable) effect).spectrum$setIncurable(false);
+                // manually remove fatal slumber to bypass turning it into eternal slumber
+                if (effect.getEffectType() == SpectrumStatusEffects.FATAL_SLUMBER)
+                    living.removeStatusEffect(SpectrumStatusEffects.FATAL_SLUMBER);
```
imreallybadatnames: ... Why? Can you not just unconditionally call that outside of the loop?

src/main/java/de/dafuqs/spectrum/mixin/compat/quilt_status_effect/absent/LivingEntityPreventStatusClearMixin.java
```diff
@@ -21,22 +23,42 @@ public abstract class LivingEntityPreventStatusClearMixin {
...
+	public abstract boolean addStatusEffect(StatusEffectInstance effect);
+
+	@Shadow
+	public abstract Map<StatusEffect,StatusEffectInstance> getActiveStatusEffects();
 	
+	private boolean hasFatalSlumber = false;
+
+	@Inject(method = "clearStatusEffects", at = @At("HEAD"))
+	private void spectrum$detectFatalSlumber(CallbackInfoReturnable<Boolean> cir) {
+		hasFatalSlumber = getActiveStatusEffects().containsKey(SpectrumStatusEffects.FATAL_SLUMBER);
```
imreallybadatnames: This is injected into the same method; why not use MixinExtras' `@Share` instead?
FYI: [this](https://github.com/LlamaLad7/MixinExtras/wiki/Share)

src/main/java/de/dafuqs/spectrum/mixin/EffectCommandMixin.java
Robotgiggle: ... not sure how i missed that lmao

src/main/java/de/dafuqs/spectrum/mixin/compat/quilt_status_effect/absent/LivingEntityPreventStatusClearMixin.java
```diff
@@ -21,22 +23,42 @@ public abstract class LivingEntityPreventStatusClearMixin {
...
+
+	@Inject(method = "clearStatusEffects", at = @At("TAIL"))
+	private void spectrum$applyEternalSlumberIfFatalSlumberRemoved(CallbackInfoReturnable<Boolean> cir) {
+		if (hasFatalSlumber) {
+  			addStatusEffect(new StatusEffectInstance(SpectrumStatusEffects.ETERNAL_SLUMBER, 6000));
+		}
+	}
+
 	@WrapWithCondition(method = "clearStatusEffects", at = @At(value = "INVOKE", target = "Lnet/minecraft/entity/LivingEntity;onStatusEffectRemoved(Lnet/minecraft/entity/effect/StatusEffectInstance;)V"))
 	private boolean spectrum$preventStatusClear(LivingEntity instance, StatusEffectInstance effect, @Share("blockRemoval") LocalBooleanRef blockRemoval) {
```
imreallybadatnames, line 46: Actually, `@Share` is literally used here, a couple lines later, so why did I even bring it up separately?

src/main/java/de/dafuqs/spectrum/status_effects/ImmunityStatusEffect.java
```diff
@@ -19,13 +22,28 @@ public boolean canApplyUpdateEffect(int duration, int amplifier) {
...
+		removeNegativeStatusEffects(entity);
+	}
+
+	public static void removeNegativeStatusEffects(@NotNull LivingEntity entity) {
+		Set<StatusEffect> effectsToRemove = new HashSet<>();
+		Collection<StatusEffectInstance> currentEffects = entity.getStatusEffects();
+		for (StatusEffectInstance instance : currentEffects) {
+			StatusEffect effectType = instance.getEffectType();
+			if (effectType.getCategory() == StatusEffectCategory.HARMFUL && !SpectrumStatusEffectTags.bypassesImmunity(effectType)) {
+				effectsToRemove.add(instance.getEffectType());
```
imreallybadatnames: nit: why do you use `instance.getEffectType()` here instead of the used-above `effectType`?

Robotgiggle: oops, missed that while bringing it over from WhispyCircletItem

src/main/java/de/dafuqs/spectrum/mixin/compat/quilt_status_effect/absent/LivingEntityPreventStatusClearMixin.java
Robotgiggle: entirely on me, this is my first time doing something this complex and I had no idea that existed

DaFuqs (commented):
Lots of good changes, thank you!

> if fatal slumber is applied, incoming immunity converts fatal->eternal and doesn't apply immunity
Why would it not be removed / prevented to be applied? That feels like a weird, inconsistent edge case that I would not expect as a player

> public static TagKey<StatusEffect> BYPASSES_WHISPY_CIRCLET;
> public static TagKey<StatusEffect> BYPASSES_NECTAR_GLOVES;
> public static TagKey<StatusEffect> BYPASSES_IMMUNITY;

Definitely a lot better named than before.
Why would we not be able to merge them into a single tag, though (whispy & immunity feel they could share the tag, no idea about nectar gloves yet)

Robotgiggle:
> Why would it not be removed / prevented to be applied? That feels like a weird, inconsistent edge case that I would not expect as a player

I'm not sure what you mean here, since this follows the exact same behavior as every other interaction between Immunity and Fatal Slumber. When Fatal Slumber is cured, it turns into Eternal Slumber. When Immunity is used to cure Fatal Slumber, the Immunity gets cancelled out too. Thus, if you currently have Fatal Slumber and then drink an Immunity potion, the Fatal Slumber will become Eternal Slumber, all other negative effects will be cured as usual, and the Immunity will not apply since it was cancelled out by curing the Fatal Slumber.

If the Immunity _did_ still apply, it would immediately cure the newly applied Eternal Slumber in the next tick, which would invalidate the whole "Fatal Slumber takes two steps to cure" thing.

> Why would we not be able to merge them into a single tag, though (whispy & immunity feel they could share the tag, no idea about nectar gloves yet)

Immunity has a separate tag because the Gloves and the Circlet are both unable to cure `#spectrum:soporific` effects, while Immunity is able to cure everything. Currently it would absolutely be possible to use a single tag for both the Gloves and the Circlet, but I decided to make separate ones just to keep things more specific. Let me know if you think I should merge those two.

DaFuqs:
> If the Immunity did still apply, it would immediately cure the newly applied Eternal Slumber in the next tick, which would invalidate the whole "Fatal Slumber takes two steps to cure" thing.

That is my main point? Why would that be a problem? Currently vanilla has no single case where an effect needs to be cleared twise, so this is not something I would expect as a player. Do we lose a lot making fatal slumber being able to be cured like every other effect? If a player has milk/clotted cream on them, they probably have more than one anyways. If not... not.

> [...] the Gloves and the Circlet are both unable to cure #spectrum:soporific effects

But why :concern:

Robotgiggle:
Taking two steps to cure is my answer to the fact that Fatal Slumber being curable at all makes it largely useless in PvP. It's _very_ hard to get your hands on Fatal Slumber in the first place, since you have to get 4 burgeons and then stick them in a titration barrel for 24 hours, rather than the much simpler Eternal Slumber recipe which just needs 1 burgeon in a potion workshop. If a single cure-step was able to get rid of Fatal Slumber entirely, there'd be no reason to use it over Eternal Slumber, since assuming the target is going to be able to cure it before it kills them (which, if you have any sort of effect cure in your hotbar, is not that hard), the actual effect of Fatal Slumber is not that different from Eternal Slumber despite the much greater cost.

On the other hand, making it entirely incurable feels unfair, as that would essentially make applying it an instant win button. Making it take two steps to cure avoids both of these problems – there's a reason to use it over Eternal Slumber, namely that it'll force your opponent to take more time than normal purging the effect, but it's also no longer a guaranteed kill.

It's definitely true that someone who has one clotted cream will probably have more, but the same argument applies to the Incurable effects made using Bitter Oils. The point isn't to make it impossible to get rid of, just to make it harder to cure than a normal effect, due to both its power and the amount of effort you have to put in to obtain it in the first place.

As for the strangeness of an effect needing to be cured multiple times, there very much is a precedent for that: the aforementioned Incurable effects. In fact, unless the duration is particularly short, one of those will likely take quite a bit more than two steps to cure. This certainly isn't a vanilla feature, but I don't think that alone is a reason to avoid it – this is a mod, after all.

PR #598, opened 2024-11-22
Bedrock Anvil Item Coloring

f-raZ0R:
Combine items with pigment in the bedrock anvil to color their names.

Minor concern i've noticed is that the act of resetting an item's name, while being consistent with vanilla, is a bit unintuitive.
I made a note of it in the guidebook, but I worry that it might not be enough.


DaFuqs:
Awesome! I tried out a few things and it seems to work great.

PR #599, opened 2024-11-22
Add quality to the loot only infused beverages that are really good

f-raZ0R:
luck could always use some more love
get better loot!


f-raZ0R:
i. why is it spaced out that way

DaFuqs (approved):

DaFuqs:
> i. why is it spaced out that way

Tab supremacy

PR #601, opened 2024-11-24
Refactor Puff Circlet fall negation logic.

imreallybadatnames:
Motivation: compatibility w/ Botania and other mods, as well as proper handling of other checks, e.g. falling into water.


DaFuqs:
❣

PR #602, opened 2024-11-24
hotfix: accidental inclusion of non-functional draft mixin

imreallybadatnames:
Revert non-functional post-test draft changes done to fix a secondary problem [namely passengers being hurt even if the vehicle entity is protected], that, ironically, reverted the actual primary solution [see prev. PR].
Gigantic brain fart on my part; didn't process the mistake until now.

### DIFF
--- a/src/main/java/de/dafuqs/spectrum/mixin/LivingEntityMixin.java
+++ b/src/main/java/de/dafuqs/spectrum/mixin/LivingEntityMixin.java
@@ -385,15 +385,15 @@ public abstract class LivingEntityMixin {
 		return (float) this.getAttributeValue(EntityAttributes.GENERIC_ARMOR_TOUGHNESS);
 	}
 
-	@Inject(method = "handleFallDamage", at = @At("HEAD"), cancellable = true)
-	private void spectrum$puffCircletDamageNegation(float fallDistance, float damageMultiplier, DamageSource damageSource, CallbackInfoReturnable<Boolean> cir) {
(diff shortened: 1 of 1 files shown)


DaFuqs:
Looks good so far.
One thing that comes to mind is that the change does make the Puff Circlet not trigger if the actual damage is smaller than the one the Trinket would use up on dike. Since the Circlet does not proc its effect it may make the players think is is broken / does not trigger in all cases. How should we handle that?

imreallybadatnames:
> One thing that comes to mind is that the change does make the Puff Circlet not trigger if the actual damage is smaller than the one the Trinket would use up on dike.

~~uhh that was the initial plan~~ We could just clamp the absorbed dike value between 0 [not including] and the dike cost [including], unless that's a no-go.

DaFuqs:
Sounds perfectly reasonable. Also updating the guidebook lang to mention it uses *max of* x dike to absorb a fall.

imreallybadatnames:
Also, there's a [rather minor] nuance that the full damage is still absorbed even if the dike didn't absorb the cost. How should that be accounted for?

DaFuqs:
Good point.
Only being able to trigger the effect is there is enough dike for the fall (either enough to negate damage, or the max cost) feels like the way to go

imreallybadatnames:
Done. Anyways, should I do the guidebook changes, or...? Seems rather cumbersome, especially considering I don't know most of the supported languages.

DaFuqs:
No prob. I will take a look at the guidebook and see what I can do

DaFuqs:
Guidebook entries & tooltip are worded in a way we do not need to change anything in the first place ✅

PR #606, opened 2024-11-26
Fix and update reagent descriptions

Robotgiggle:
- Unswaps the listed potency modifiers for Raw Malachite and Raw Bloodstone
- Rewords a few other reagent descriptions
- Changes all reagent page lang to use `book.spectrum.guidebook.potion_workshop_reagents.<REAGENT_NAME>.text` rather than using (often incorrect) page numbers for half of the entries and names for the other half


DaFuqs (approved):
Awesome!

PR #610, opened 2024-11-28
Improve special-case amalgam explosions

Robotgiggle:
- Eating amalgam no longer reduces the block damage of the resulting explosion
  - This was not documented anywhere, and led to some players being confused why their amalgam wasn't able to crack dragonbone
- Explosion when eating amalgam is now centered on your head rather than at your feet
  - This both makes logical sense (why would eating an explosive make your feet explode) and makes it a bit easier to actually hit things with the explosion
- Reworked explosion power formula for damaging an in-world stack of amalgam
  - Explosion power from a damaged stack is now `base power * [ 0.875 + (stack size / 8) ]`
  - A stack of 1 item will explode identically to the placed or eaten version
  - Each additional item in the stack will add 0.125 to the explosion power multiplier
  - A stack of 2 items will explode with 1.125x power, 3 items will explode with 1.25x power, etc


DaFuqs:
> Eating Amalgam to crack Dragonbone

Okay then

DaFuqs (approved):
Great changes. Ty!

PR #615, opened 2024-11-30
Make survive_fatal_slumber trigger from surviving in any way

Robotgiggle:
Currently, the trigger for the "survive fatal slumber" advancement is only activated if you survive by passing the isImmuneish() check that reduces the damage from Float.MAX_VALUE to 95% of your max health. This changes it so the advancement triggers if you survive the kill shot in any way.

Note that "in any way" does not include using a Totem of Undying, since for some reason Fatal Slumber damage is set to bypass that.


Azzyypaaras:
It bypasses it because it would be stupid

PR #619, opened 2024-12-01
Automatically migrate pre-InkColor registry InkStorages

unilock:
These changes allow `InkStorage` instances created prior to Spectrum v1.8 to migrate to the new Registry-based `InkColor` system automatically.

BlockEntities and ItemStacks load as expected; the player doesn't have to do anything special.

BlockEntities will be saved with the new format whenever the game usually saves BlockEntities (e.g. upon stopping the server), and ItemStacks will be saved with the new format whenever their `InkStorage` is modified (e.g. upon adding or removing ink).


DaFuqs:
👌

PR #629, opened 2024-12-05
Remove "brew a vanilla potion first" from potion unlocks

Robotgiggle:
For some reason, the unlocks for all of Spectrum's modded potion recipes require you to first brew a vanilla potion in the potion workshop. I can't see any reason why this would be necessary, and it seems to me that it just adds an extra step to the unlock process that's both tedious for experienced players and non-obvious for new players. Thus, this PR removes that requirement from all the potion recipes.


DaFuqs:
At the moment I do not vibe with this suggestion.
I feel just throwing it all at the player at once when unlocking the Potion Workshop would be a tad much for a first time player and also lack the additional wow effect from having tried it out.

When first unlocking the Potion Workshop, the player has a *lot* to read through, the new brewing gui, all about the reagent system, how it works, all the different reagent affects and everything else the Potion Workshop is able to craft. - especially for players that just skim through the book that is a lot already.

If we remove that check we might end up in a different, yet similar situation: "Oh, I did not realize there are effects that are not vanilla!". I just read the first page / clicked "mark all as read".

Right now the unlock condition gives the player a few minutes of time to get accustomed to this new way of brewing using recipes that are known to them and only after that gets the unlock popup with "oooooh, I can do even more!". It also has flavor in the sense of the player realizing what else the Workshop is capable of.

Robotgiggle:
That's a fair point, but the issue with the current system is that, when a new player unlocks the workshop and sees that they can only make vanilla potions, they may just never bother trying it (since vanilla potions aren't normally very useful in modded). If this happens, they'll never find out that there's new things they can make. 

This is particularly relevant for the new sleep effects, since the potion workshop recipes are the only way to access them in a form that isn't only applicable to yourself. It seems strange to me that a player could complete all the other requirements (ie eating nightdew/burgeons) and/or even obtain the Poisoner's Handbook, but still be unable to brew the sleep effects since they've never made a vanilla potion - especially since nothing indicates that doing so is necessary to unlock the sleep recipes.

DaFuqs:
I do not know what the best solution to this is yet, but removing these unlock conditions does not feel like it.
Ideally we'd give the players more incentives to tinker with the Workshop in the first place.

PR #630, opened 2024-12-05
changed vein mining max level text

800020h:
I haven't tested it


DaFuqs:
That is a tricky one. Even spectrum's own enchantments have a configurable max level.
For consistency we would have to add that disclaimer to every spectrum enchantment's guidebook entry, too.
But that will clutter the book. Ideally we would change that text dynamic based on the values in the config.

I've created a feature request for Modonomicon here: https://github.com/klikli-dev/modonomicon/issues/274

PR #632, opened 2024-12-05
fix: server explode due to CCA trickery

imreallybadatnames:
On an unrelated note, my commits now show up as Unverified, despite my Environment configured in a way that it *requires* said keys to even push. Strange.


imreallybadatnames:
... They're Verified now??? The GitHub UI is truly fascinating.

imreallybadatnames:
Wait a minute. All these commits weren't there before! What is going on!?
... Whatever, the resulting changes are only the fix code.

PR #640, opened 2024-12-14
Updated Chinese translations

EndlessLeo:
Added Chinese translations of source ideas, the unbinding page, and the new advancement. Fixed some translation errors in advancements.


DaFuqs:
thanks!

PR #642, opened 2024-12-21
Lootr compatibility for Amphoras

unilock:
This PR adds compatibility with [Lootr](https://modrinth.com/mod/lootr) for Amphoras that contain loot, generated as part of structures.

TODO:
- [x] Block models, blockstates, translations (en_us), etc.
- [ ] Block textures (if separate textures are desired)
  - [ ] Different textures for looted and unlooted Amphoras (a la [`BarrelModel.java`](https://github.com/LootrMinecraft/Lootr/blob/fabric-1.20/src/main/java/net/zestyblaze/lootr/client/block/BarrelModel.java))
- [x] ~~Config option to toggle this (as Lootr has for its blocks)~~ (just adding "lootr" to "IntegrationPacksToSkipLoading" should have the same effect)
- [x] ConfigManagerMixin to automatically replace generated Amphoras with their Lootr versions
- [ ] Custom ScreenHandler UI (currently just looks like a vanilla double chest named "Amphora" - plain background, left-aligned title, etc.)
- [ ] Lootr compatibility for Weeping Gala Barrel (Preservation Chests seem to use similar logic already)
- [x] Testing!!!

This will not build until this Lootr PR is merged (and released): https://github.com/LootrMinecraft/Lootr/pull/557


DaFuqs:
Since 1.21.1 is the current dev target and this is still WIP, I'll close it for now. Would be a cool feature to have!

unilock:
Unfortunately this particular PR will probably never be ready to merge, as Lootr *still* hasn't made a new release with the fixed mappings.

However, if I remember correctly, Lootr's API is greatly improved on 1.21, so perhaps it'll be an easier task to port up :)

PR #649, opened 2024-12-29
Avoid usage of Local in AbstractInventoryScreenMixin

unilock:
Fixes a weird crash during startup in AOF7. Tested and working. (well, it works the same as upstream Spectrum, anyway)


imreallybadatnames:
yummers

PR #653, opened 2025-01-01
Assorted bug fixes

Robotgiggle:
- Blood orchid petals now unshroud from either solving a wireless redstone ruin or finding an undergrowth manor
- New tag `cannot_be_incurable` for effects that don't work with Bitter Oils (by default, contains Eternal and Fatal Slumber)
- When brewing an effect that's in `cannot_be_incurable`, trying to add Bitter Oils will now make the recipe invalid
- Soothing Bouquet hint is now "Perform at the darkest hour" rather than "Perform at the height of slumber"
- Bristle Sprouts, Varia Sprouts, and both types of Moss Ball can now be harvested using Silk Touch as well as shears
- Various gala wood products now work as furnace fuel, and burn twice as long as normal wood
- Mentioned in the guidebook that gala wood is really hard and burns more slowly than average
- Botania petal blocks can now be used in the color picker, for 9x the value of a petal
- Fixed the place where you could sneak into the manor labyrinth through the basement floor
- Enchanter cost calculation now properly reduces level scaling for enchants with higher max levels


DaFuqs:
Looks great!
The only thing that concerns me more and more are all the edge cases the slumber effects need that are inconsistent to other statuus effects and hard to communicate to the player. But without a big refactor there is nothing much that can be done about that.

Issue #656, opened 2025-01-13
[BUG] Slate noxwood fence copy does not exist


jessethomascs:
Issue: You cannot craft slate noxwood fences in the game by default
Version: 1.20.1

Reason: The json file from slate_noxwood_fence seems to have been copied-pasta from slate_noxwood_fence! so the files are identical, and the fences cannot be crafted in survival

Sources:
FENCE: https://github.com/DaFuqs/Spectrum/blob/1.20.1-aria-for-painters/src/main/resources/data/spectrum/recipes/crafting_table/noxwood/slate_fence.json

GATE: https://github.com/DaFuqs/Spectrum/blob/1.20.1-aria-for-painters/src/main/resources/data/spectrum/recipes/crafting_table/noxwood/slate_fence_gate.json

jessethomascs:
the fix for this is just adjusting the single json file too AFAIK

jessethomascs:
https://github.com/DaFuqs/Spectrum/pull/657 Created a fork for this patch

PR #657, opened 2025-01-13
[1.20.1-noxwood-fence-patch] Patch noxwood shaped crafting bug

jessethomascs:
Fix copy-pasta to left users craft noxwood fences


jessethomascs:
Addresses https://github.com/DaFuqs/Spectrum/issues/656

PR #658, opened 2025-01-14
Pastel Network Rearchitecturing (BE graph -> BlockPos graph)

Azzyypaaras:
*Go on.*
*Cry to your mama, eat your crayons.*
*You can't  un-piss the pastel network.*

image

This PR overhauls the internals of the pastel network to fix various issues with it - most notably the extremely problematic server freezes that have been an urgent issue.

To achieve this, a substantial internal overhaul was required. Of note for the major changes are:

- The internal _graph_ of the networks are now based on BlockPos objects instead of BlockEntities, this allows the graph to be serialized instead of needing to be reconstructed on the fly.
- There is now a differentiation between the vertices of a network and its loaded nodes. A node need not be loaded for a network to consider it a part of itself and be valid for pathfinding. This is probably has positive effects for performance.
_ Common code between Server and Client components of networks has increased.
- The Server and Client components of networks are now kept more closely synchronized. Ahoy sailor be wary, tharr be janky hacks 'ere.

As a very pleasant side-effect of this overhaul, pastel nodes are now able to transfer items over unloaded chunks as long as the transmitter and receiver are loaded.

Further stress-testing before release would be good. While I am fairly certain everything works as intended, some extra testing is always good.

<sub>Footnote: This ordeal has only reinforced my hatred for **PersistentState**</sub>


src/main/java/de/dafuqs/spectrum/blocks/pastel_network/network/ClientPastelNetworkManager.java
```diff
@@ -52,58 +53,62 @@ public Optional<? extends PastelNetwork> getNetwork(UUID uuid) {
...
 	public void removeNode(PastelNodeBlockEntity node, NodeRemovalReason reason) {
 		PastelNetwork network = node.getParentNetwork();
 		if (network != null) {
 			network.removeNode(node, reason);
-			if (network.nodes.size() == 0) {
+			if (network.graph.vertexSet().isEmpty()) {
 				this.networks.remove(network);
 			}
+			else {
+				checkForNetworkSplit(network);
```
DaFuqs: Why is network splitting also checked for on the client side? o.O

src/main/java/de/dafuqs/spectrum/blocks/pastel_network/network/PastelNetwork.java
```diff
@@ -338,6 +269,77 @@ public String getNodeDebugText() {
                 " - Conn: " +
                 getNodes(PastelNodeType.CONNECTION).size();
     }
+	
+	public NbtCompound toNbt() {
```
DaFuqs: NBT is implemented for both server & client networks? o.O

src/main/java/de/dafuqs/spectrum/blocks/pastel_network/network/ServerPastelNetwork.java
```diff
@@ -73,6 +67,8 @@ else if (transferLooper.getTick() % 2 == 0) {
 		if (cap || priority != Priority.GENERIC) {
 			if (cap) {
 				this.transferLooper.reset();
+				// If someone complains about network usage I am going to fucking kill myself
+				SpectrumS2CPacketSender.syncPastelNetworkEdges(this, toNbt());
```
DaFuqs: But what about network usage??

src/main/java/de/dafuqs/spectrum/blocks/pastel_network/network/ServerPastelNetworkManager.java
```diff
@@ -50,12 +55,18 @@ public NbtCompound writeNbt(NbtCompound nbt) {
 	public static ServerPastelNetworkManager fromNbt(NbtCompound nbt) {
 		ServerPastelNetworkManager manager = new ServerPastelNetworkManager();
 		for (NbtElement element : nbt.getList("Networks", NbtElement.COMPOUND_TYPE)) {
-			manager.networks.add(ServerPastelNetwork.fromNbt((NbtCompound) element));
+			var compound = (NbtCompound) element;
```
DaFuqs: TODO: look into this

src/main/java/de/dafuqs/spectrum/blocks/pastel_network/network/ServerPastelNetworkManager.java
```diff
@@ -154,52 +159,4 @@ public void removeNode(PastelNodeBlockEntity node, NodeRemovalReason reason) {
 		}
 	}
 	
-	private void checkForNetworkSplit(ServerPastelNetwork network) {
```
DaFuqs, line 157: TODO: why was this moved away from the server side?

DaFuqs:
It arrived, ty!
I'll go through everything and leave my WIP notes on the way.

src/main/java/de/dafuqs/spectrum/blocks/pastel_network/nodes/PastelNodeBlockEntity.java
```diff
@@ -55,9 +55,10 @@ public class PastelNodeBlockEntity extends BlockEntity implements FilterConfigur
 
 	@Nullable
 	protected PastelNetwork parentNetwork;
+	@NotNull
```
DaFuqs, line 56: TODO: what?

src/main/java/de/dafuqs/spectrum/blocks/pastel_network/nodes/PastelNodeBlockEntity.java
```diff
@@ -438,9 +417,19 @@ public NbtCompound toInitialChunkDataNbt() {
 	@Override
 	public void markRemoved() {
 		super.markRemoved();
-		Pastel.getInstance(world.isClient).removeNode(this, NodeRemovalReason.UNLOADED);
+		// Hanky jacks
```
DaFuqs: Handy jacks? But I think I already see why

src/main/java/de/dafuqs/spectrum/networking/SpectrumS2CPacketReceiver.java
```diff
@@ -543,7 +544,19 @@ public static void registerS2CReceivers() {
 				}
 			});
 		}))));
-
+		
+		ClientPlayNetworking.registerGlobalReceiver(SpectrumS2CPackets.PASTEL_NETWORK_EDGE_SYNC, (client, handler, buf, responseSender) -> {
```
DaFuqs, line 549: TODO: can this be handled in the node block entity sync packet instead?

DaFuqs (commented):
Lots of great improvements all around.
I'll run tests ingame and will look at why some things are the way they are. I am sure you had good reasons.

src/main/java/de/dafuqs/spectrum/blocks/pastel_network/nodes/PastelNodeBlockEntity.java
```diff
@@ -55,9 +55,10 @@ public class PastelNodeBlockEntity extends BlockEntity implements FilterConfigur
 
 	@Nullable
 	protected PastelNetwork parentNetwork;
```
DaFuqs: Why is that property called *PARENT*network?

imreallybadatnames: just cuz

DaFuqs:
1. Adding a new node to a network changed the color of the whole network. IMO it should preserve the network UUID of the bigger of the previous networks to not change color all the time
2. Having a Provider, Buffer and Gather node in the same network made items travel from Provider => Buffer => Gather, instead of Provider => Gather directly, despite there being room (no idea it was like that before already)

DaFuqs:
Other than the things I found works perfect!

src/main/java/de/dafuqs/spectrum/networking/SpectrumS2CPacketReceiver.java
DaFuqs: On second thought, I like your approach a lot

DaFuqs:
The unpissing shall be real

DaFuqs:
Inspired by your idea to sync network states I changed up the clientside to only ever store the network UUID and its graph. Literally everything else (manipulation of nodes, ticking, connections, ...) are all handled purely serverside now. the client classes are mostly mere holders for rendering now.

All changes on the server side that modify the graph are synced to the clientside. No more worries to keep everything in sync manually! Saves >100 loc and headaches. Network traffic has been drastically reduced.

DaFuqs:
Oh yeah, the `PastelNodeBlockEntities` were changed a bit, too:
- They do not hold a reference to their network anymore which prevented those from being garbage collected before
- They keep their internal ID, so whenever you unlink & relink them, they use the same network color as before

TODO: ~~when joining a world the nodes do not render yet. Fully functional, though, since all logic is ran serverside.~~ DONE

Azzyypaaras:
I took a look through things to make sure things works well, and everything looks great! Nothing popped up past a minor visual desync with upgrades - removing an upgrade will have its visuals remain on the node until rejoining the world.

I gave fixing it a short try but decided that you probably have a better idea of how to handle this since upgrades are in a bit of a middle spot when it comes to the server and client.

That aside, to unwind from the pastel pissing and also because I have lost control of my life and I just automatically fiddle with rendering whenever I get within 300 feet (omg feet) of client code. I made some minor visual improvements for nodes; the pastel network is a bit more colorful now.

DaFuqs:
The node upgrade code is kind of cursed:
- Reading the nodes nbt (like loading a chunk) will immediately always mark it as dirty again
- If a node has an upgrade and nbt specifies it has none, the original upgrade is getting kept and not cleared
- upgrade adding runs server- & clientside, but removing it only serverside

PR #665, opened 2025-02-03
Advancement typo fixes

KatTheFox:
it's


Azzyypaaras:
Statistically this is probably all correct so

PR #671, opened 2025-02-09
Potion unlocking rework (v2)

Robotgiggle:
A couple months ago, I made a PR (#629) to remove the requirement for brewing a vanilla potion from all of the spectrum potion recipes. It was denied, on the grounds that dumping the entire Potion Workshop system on new players at once would be overwhelming, which is a fairly valid point.

However, the problem that PR would have solved is still a thing, and we're still getting people in the questions channel who've beaten the entire main progression but don't know how to unlock potions. After discussing this issue with a couple other devs (conversation starts [here](https://discord.com/channels/906264467636109333/906274936417554464/1337915018737487883)) I believe we've come up with a solution that fixes the "how do I unlock potions?" issue without making the system too overwhelming for newbies.

This PR removes the brew-a-potion requirement from the spectrum potion recipes, just like the previous one did. However, it also _adds_ that advancement as a requirement for all of the custom reagents added by Spectrum. After all, the reagent system is the most complex part of the Potion Workshop's mechanics, so if we're trying to reduce information overload that seems like the best thing to temporarily lock off.

When the Potion Workshop is first unlocked, players will have access to all of the custom spectrum potions whose ingredients they've unlocked, and the reagents entry will contain only the 3/4 vanilla ones (redstone, glowstone, gunpowder, and potentially dragon's breath). The opening pages of the entry will look like this:
image

Once the player brews a potion using the Workshop, all of the custom reagents added by Spectrum will unlock. This will also cause page 2 to appear at the beginning of the reagents entry, describing how the new reagents can interact with each other:
image

This way, the flood of information for new players is significantly reduced, _without_ the current issue of removing any incentive to actually use Spectrum's potion mechanics and thus discover the rest of the content.

This PR also makes locked reagent recipes actually act locked. Previously, if a reagent recipe was locked it would display as such in your recipe viewer, but putting the item into the Potion Workshop would work totally fine.


DaFuqs:
Big fan!

With that we also would have the option to unlock the potion workshop earlier, since the power level is a lot lower at the start now, giving quitoxic a new early game unlock. I'll note that as a possible improvement down the line.

PR #674, opened 2025-02-18
Various fixes

Robotgiggle:
- The fusion shrine now works again
- Blackest Lotus now produces exactly 1 vanilla pool's worth of mana instead of using the max capacity of the pool (prevents cheese with mods like Applied Botanics)
- Azalea Tea's secret milk recipe now has the same "Be a tea philistine" hint as the other milk tea recipes
- Fixed up another hole in the manor labyrinth
- A variety of Spectrum subtitles are now more closely matched to the vanilla format
- Shale and its variants now properly detect skylight for weathering purposes
- Ruin and Ruined Bedrock are now in `#c:relocation_not_supported`, along with all other Spectrum unbreakables
- Spectrum's custom trinket slots now use the same icon color as normal slots
- Dragon Talon now correctly accepts sword enchantments rather than trident enchantments
- Light Blue Lamp recipe now has the correct unlock (previously it used the `light_grey_lamp` unlock, presumably a typo)
- Staff of Remembrance now properly checks for the `unlock_hostile_memorizing` advancement
- `An obervation` -> `An observation`


Azzyypaaras:
Per chance, does the glove change in any way allow you to use two types of spectrum glove at once.

Just asking because they are not at all balanced around that

Robotgiggle:
This does allow you do use both gloves at once, yes. I was assuming the fact that you couldn't was an oversight.

Robotgiggle:
Would it be more balanced if I kept the "gloves can go in either slot" effect but made it so Spectrum alone only provides a single glove slot? That way, you'd only be able to use both of them in situations where another mod is providing the other slot, in which case there will probably be at least one other glove item you'd have to give up on to use both Spectrum gloves.

Robotgiggle:
Reverted the glove change, created emilyploszaj/trinkets#368 instead

Azzyypaaras:
- You can now apply Loyalty to the Draconic Twinsword
?

DaFuqs:
lots and lots of good stuff. Thank you!

PR #681, opened 2025-02-28
Remove redundant codes

Ethanolgie:
These codes will never be excuted in game because if age == 2,hasRandomTicks = false.


DaFuqs:
that is definitely for the best. Randomly dropping items isn't particularly good for performance

PR #686, opened 2025-03-09
Custom slot for spectrum gloves

Robotgiggle:
Unlike most glove trinkets added by other mods, Spectrum's gloves each represent a pair of gloves rather than a singular glove, and thus Spectrum only enables one of the two glove slots to put them in. This causes jank when another mod that enables both glove slots is installed, as the Spectrum gloves can only go in one of the two slots with no immediate indication as to why.

This PR solves the issue by creating a custom `gloves` slot for Spectrum's gloves to go in. Now that the gloves have their own dedicated slot, there will no longer be any confusion as to why they can only fit in one of the two identical glove slots used by most other mods, and their mutual exclusivity with each other is still maintained.


DaFuqs:
That definitely solves the issue of being able to not equip spectrum gloves in the second glove slot. On the other hand it enables the player to equip spectrum gloves + two gloves by other mods, so they now wear three gloves with two hands.

Not sure that is necessarily better - maybe worse. Hmmmm

DaFuqs:
What if we sneakily renamed the glove slots to `Left Glove` and `Right Glove`, respectively? That way the player could immediately see that the Spectrum glove(s) can only be equipped in one of them.

Robotgiggle:
That would work, but I'm not sure how we'd go about changing the lang inside Trinkets - would that require a mixin, or can we just override it somehow?

Also, not a huge issue, but it should probably be offhand/mainhand rather than left/right since there's a settings option to use your left hand as the main one.

DaFuqs:
Since afaik the load order in fabric is mod id dependent (impl detail and should not be relied upon), `spectrum` will load before `trinkets`, so a simple lang file addition won't help here.

So yes, we would either need to bundle an always active resource pack (cluttering the players resource packs screen), or use our existing i18n mixin to inject that.

DaFuqs:
Renaming the slots in Trinkets isn't really feasible, since those seem to be connected in "trinket groups".

Because of that, I implemented another variant to resolve your issue: The spectrum gloves have a modifier of `-1 offhand glove slot`.

As a result, when another mod enabled the offhand glove slot and the player equips one of the spectrum trinkets, their offhand glove slot vanishes (or their count reduced by one, if they have multiple). Should there be an item in that slot already it is put into the players inventory.

Robotgiggle:
Ah nice, that neatly resolves the issue. Didn't realize there was a way to dynamically change slots based on equipped items like that.

DaFuqs:
Ring of multiple Rings when?

PR #687, opened 2025-03-10
Change seed vault disc id to new version id and add some Exclusions Lib predicate for geodes

Ethanolgie:
1.8.0 changed disc id into music_disc_credits,while the disc id in seed_vault.nbt is still music_disc_dimension_theme,so no disc will actually be in that jukebox.


Ethanolgie:
image
image
image
Well...This is why.

PR #696, opened 2025-03-24
Add Create train hat info for some spectrum mobs

Pyredrid:
In the Create mod, entities are given hats when made into stock keepers and scheduled train drivers.  These JSONs contain info for how those hats should be parented and positioned for kindlings, erasers, and lizards.

Also I did not want to deal with setting up a new key on github so I did these entirely through GH's "Add new file" prompts.  Someone please test this before pulling, lmao.


Pyredrid:
Oh its worth noting that egg laying wooly pigs seem to not get create hats at all for some reason.  Hence why they aren't included here.  I suspect its something to do with them already having a hat model of some sort?  I didn't dig into it too far.

Pyredrid:
Heck yea!  :rat:

PR #704, opened 2025-04-09
Potion unlock rework, 1.21 edition

Robotgiggle:
All of the changes from #671, but for the 1.21 version. 

Currently a draft, since I haven't actually tested any of this apart from making sure it builds.


DaFuqs:
Could you get it to run?

Alternatively I can have a look no problem. IIRC the potion unlock rework here is the last thing to get all 1.20.1 and 1.21 up to par.

Robotgiggle:
Should be good to go now.

DaFuqs (approved):
very, very nice. Ty!

PR #705, opened 2025-04-09
Bugfix pr, 1.21 edition

Robotgiggle:
All of the changes from #674, but for the 1.21 version. Does not include the TranstargetItem fix or the manor labyrinth fix because those are already handled.
In addition, fixes the missing lang key for the Bismuth Trinkets book entry (at least in English) and corrects a typo in the section describing the DD's downstone floor.

Currently a draft, since I can't test anything until I figure out how to make 1.21 Spectrum actually run on my machine.


DaFuqs:
Can you include #671 too, pls?

Robotgiggle:
That's what #704 is - or do you want me to put them together into a single PR?

DaFuqs:
Ah, my bad. Totally missed that one

Noaaan:
The seperate PR is fine. I'll merge this once the conflict is fixed

PR #709, opened 2025-04-26
Many simple fixes

Ethanolgie:
Update Draconic twinsword unlock condition because you can reveal blood orchid petals upon reaching Undergrowth manor this version.However,you cannot craft twinsword unless you had solved bedrock and redstone puzzle ruin.


Ethanolgie:
It seems that Ubuntu could not run check while Windows succeeded...

Ethanolgie:
Fixed more problems in zh_cn.json,and make pure alcohol+chrysocolla able to superheat Create blaze burner

DaFuqs (approved):

PR #711, opened 2025-04-28
Allow putting the Sweet Pea, Apricotti, Humming Bell and Varia Sprout in flower pots

Auroali:


DaFuqs:
Thank you!

TODO on my side: Check if we need the separate potted flower textures, or if we can pull of some form of trickery

DaFuqs:
> Check if we need the separate potted flower textures, or if we can pull of some form of trickery

Yes, yes we do

PR #724, opened 2025-05-23
spectrum enchants are mysterious and swagful again

TheTrueYak:
-reimplemented enchantment obfuscation
-reimplemented idwtialsimmoedm compat

### DIFF
--- a/src/main/java/de/dafuqs/spectrum/SpectrumClient.java
+++ b/src/main/java/de/dafuqs/spectrum/SpectrumClient.java
@@ -4,6 +4,7 @@ import de.dafuqs.revelationary.api.advancements.*;
 import de.dafuqs.revelationary.api.revelations.*;
 import de.dafuqs.spectrum.compat.*;
 import de.dafuqs.spectrum.compat.ears.*;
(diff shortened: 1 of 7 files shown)


src/main/java/de/dafuqs/spectrum/SpectrumClient.java
```diff
@@ -84,6 +85,11 @@ public void onInitializeClient() {
 			logInfo("Registering Ears Compat...");
 			EarsCompat.register();
 		}
+		
+		if (FabricLoader.getInstance().isModLoaded("idwtialsimmoedm")) {
```
DaFuqs, line 89: That can be moved to the `SpectrumIntegrationPacks` class, same with the `ears` compat above, really. I'm assuming we put them here before `SpectrumIntegrationPacks` even was a thing. I will move them over in a separate commit ✅

DaFuqs (approved):
Fancy way to solve that!

It also gave me an idea how to possibly introduce advancement locked enchants - no idea if we want to have that again, since it causes a bit of issues with non-player tool usage, but we'll see

PR #725, opened 2025-05-25
Use StatusEffectInstance.amplifier getter method + update build workflow

unilock:
`StatusEffectInstance.amplifier` is private by default, so it's likely that Spectrum compiles against a mod that widens its access - but not transitively, hence this causing a crash when reached during runtime.

The build workflow has also been updated, as it looks like GitHub / Gradle have finally deprecated `gradle/wrapper-validation-action@v1`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,20 +18,15 @@ jobs:
     
     steps:
       - name: checkout repository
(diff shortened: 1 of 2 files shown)


DaFuqs:
Thank you! As always

PR #728, opened 2025-05-30
Fix life drain recipe unlock

Robotgiggle:
The recipe for the Life Drain potion (and the associated book page) is currently non-unlockable due to a misspelled unlock advancement. This PR fixes that.


DaFuqs:
Ty!

PR #730, opened 2025-06-09
Connector patch

eehunter:
Another patch for Sinytra Connector compatibility. While Spectrum itself has only one conflict I have found so far, there are a couple among its dependencies (including one caused by an issue in Forgified Fabric API), which I am working on fixing. To run Spectrum on Connector for 1.21.1 as of now, my fork of FFAPI is required, in addition to a NeoForge port of Exclusions Lib which I am in the process of making (I have something that works currently, but it _only_ runs on Connector at this time).
There is at least one graphical bug, which likely has something to do with Fractal. The background of the icon for the Spectrum creative tab appears to be upside down when the tab is on the top row (I have not tested it with the tab on the bottom row).


DaFuqs:
Thank you lots!

PR #737, opened 2025-06-21
Fixing prolonging not prolonging effect as expected.

Ethanolgie:
In Minecraft,the Amplifier of an effect is actually its level-1,making this will return originalDuration while you only have a level 1 prolonging...


DaFuqs:
True, true. Thank you!

PR #742, opened 2025-07-01
fix: Bottomless Bundle annoying dupe

EgrorBs:
On current version you can convert any insertable in bottomless bundle item into any other insertable item
For example convert dozens of dirt into diamond blocks
Or accidently put your fully upgraded glove into bundle with dirt, and get one more dirt...

Changes:
- Bundle now checks that it can consume itemstack type before addiung it to botomless stack

Build:
You can check build [on separate branch in gh actions](https://github.com/EgrorBs/Spectrum/actions/runs/15999889629)


DaFuqs:
thx for the info! I did a very similar change to yours which fixed the issue you described

PR #743, opened 2025-07-01
fix: Crash when there is empty tablet in fabrication chest

EgrorBs:
On current version when you put empty (without recipe) crafting tablet in fabrication chest game crashes

Changes:
- Now fabrication chest class validates `CraftingTabletItem.getStoredRecipe(...)` return value to be `!= null`

Build:
You can get ci build on [separate branch in gh actions](https://github.com/EgrorBs/Spectrum/actions/runs/15999775824)


DaFuqs (approved):
thanks!

PR #749, opened 2025-07-07
Fix pastel network

EgrorBs:
- Information about the pastel network graphs is not saved/loaded when saving the world/chunk
- Nodes don't call `initializeNode` on loading
- When painting a node, a pastel network may be created with an already existing UUID (with old color, and attached nodes), which leads to the node's malfunction after painting/repainting
- Pastel node painting only works with slingshots, which I don't think was the intended design
- Pastel node colors are not saved

Changes:
- Graph information is now saved and loaded from spectrum_pastel_network_manager.dat (as intended, I believe)
- Node calls initializeNode as soon as it ticks for the first time (I couldn't find a reliable onLoad() method)
- Network UUID is now generated randomly
- Pastel node painting now works with either paintbrushes or pigment bottles
- Pastel node colors are now saved with the BlockEntity NBT data.
- Paintbrush's tryColorBlock() now returns true when a ColorableBlock is successfully painted (as intended, I believe)


DaFuqs (approved):
Looks good, thank you!

I'll give it another round of tests just to make sure

PR #750, opened 2025-07-09
fix: Disable 'smart' culling when looking at spyhole

### LINKED ISSUE #746: [BUG] Universal Spyhole can see almost nothing underground
**Describe the bug**
I can see almost nothing underground through Universal Spyhole.

**To Reproduce**
Place the Universal Spyhole on the ground, then watch the ground through it. The screenshot here shows what is actually seen through the spyhole.
Image

**Expected behavior**
Things underground can be seen as if the player was in spectator mode and entered the ground. The screenshot here is captured in spectator mode, at same x/z as above, and 2 blocks lower.

Image

**Minecraft version**
1.21.1

**Mod version**
1.9.1

**Screenshots**
See above.

**Crash Report**
No crash.

**Notes**
This can be reproduced in 1.8.X, with Sodium installed. Not tested in 1.8.X without Sodium.
Here is my mod list in 1.9.1:

Image

EgrorBs:
Implementation feels suboptimal, but I can’t think of a better way

This should partially resolve #746, but only when you’re looking directly through a spyhole (or shortly afterward)

image
image

### DIFF
--- a/src/main/java/de/dafuqs/spectrum/registries/client/SpectrumClientEventListeners.java
+++ b/src/main/java/de/dafuqs/spectrum/registries/client/SpectrumClientEventListeners.java
@@ -67,6 +67,26 @@ public class SpectrumClientEventListeners {
 		registerCustomItemRenderer("omni_accelerator", SpectrumItems.OMNI_ACCELERATOR, OmniAcceleratorItem.Renderer::new);
 		
 		WorldRenderEvents.START.register(context -> HudRenderers.clearItemStackOverlay());
+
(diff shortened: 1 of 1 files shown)


DaFuqs:
Thank you!

That's actually a pretty straightforward way to resolve the culling getting better and better with each MC update. Disabling culling entirely when looking at a Spyhole makes it work even better than in 1.20.1 before.

One small thing that improve the user experience a lot is some kind of indicator that the culling disabling is only active while looking directly at the block. Particles or a sound effect should do the trick just fine. I'll see which works best.

Issue #460, opened 2024-08-18
[BUG] /reload causes Enchantment Upgrades to fail, requires world restart to fix.


SuperKnux:
**Describe the bug**
If you use /reload or change datapacks, Enchantment Upgrade recipes fail, becoming unable to be viewed and used. This requires a world restart to fix.

**To Reproduce**
Enter a world.
Use /reload.
Check a non-level-1 enchantment book's recipe or check the Spectrum book's enchantment category.

**Expected behavior**
The Enchantment Upgrade's recipe should be available and useable.

**Minecraft version**
1.20.1

**Mod version**
1.7.16

**Screenshots**
![image](https://github.com/user-attachments/assets/b01e041b-9343-49a1-8453-2b1e7429182b)
This doesn't normally happen.
![image](https://github.com/user-attachments/assets/83e0d7b4-b543-4eb9-b03c-da7d3faa14f1)
These appear when a /reload occurs.

**Crash Report**
If it is a crash bug add your crash log here

DaFuqs:
Related to https://github.com/KubeJS-Mods/KubeJS/issues/293
I guess the workaround I added for KubeJS broke somehow.

f-raZ0R:
Seems to have been fixed on KubeJS's side
(again, do not have the perms to close this issue myself)

Mouy07:
This is still happening on the latest version, even with no other mods installed (other than dependencies and EMI).

unilock:
I can confirm that this still occurs on Fabric 1.20.1 + Spectrum 1.8.6 with no other mods installed, aside from Spectrum, its dependencies, and EMI.

PR #762, opened 2025-08-31
Revert KubeJS compatibility changes

unilock:
Fixes enchantment upgrade recipes disappearing upon reload (#460)

Tested and working with KubeJS - I guess they stopped doing whatever horrible thing they were doing to break this in the first place


DaFuqs:
https://github.com/KubeJS-Mods/KubeJS/blob/2001/common/src/main/java/dev/latvian/mods/kubejs/core/mixin/common/RecipeManagerMixin.java

The logic definitely changed from last time I looked at it.

There still is a cancel, but only a conditional one, now. Did you test with and without custom recipes added? (or whatever registers `ServerEvents.RECIPES`? If it works even with that I have no issues ditching the current ugly workaround.

unilock:
I don't think I need to test it to know that using KubeJS to modify recipes will break things. I am more than somewhat tempted to mixin into KubeJS out of spite. But I guess it would be better to just implement `KubeJSPlugin#injectRuntimeRecipes` in a mod compat class at this point.

DaFuqs:
> I am more than somewhat tempted to mixin into KubeJS out of spite

mood

DaFuqs (approved):

DaFuqs:
Since KubeJS did never release for fabric 1.21.1+ we don't have to worry about that (for now). I'll keep an eye out regardless.

I'll merge this and create a release real quick. Thank you!

PR #763, opened 2025-08-31
prevent ring gravity effects while creative flying

Auroali:
this fixes an issue that would cause the ring of aetherial grace to make you slowly rise into the air while creative flying in survival through other mods


DaFuqs:
Thank you!

PR #765, opened 2025-09-09
Fix enchanter cannot enchant items

### LINKED ISSUE #764: [BUG] Enchanter cannot enchant items with enchanted books
**Describe the bug**
The enchanter cannot enchant items. When a tool is place in the center and an enchanted book is placed in one of the item bowls, XP in the knowledge gem is consumed, but item cannot be enchanted.

**To Reproduce**
1. Construct an enchanter structure;
2. Place an knowledge gem with 10000 XP inside in the center;
3. Place an enchanted book (with enchantment Infinity) in one of the item bowls;
4. Place an bow (without enchantments) in the center;
5. Knowledge gem starts to create particles, purple points starts to fly into the enchanter indicating that enchanting has started. **But there is no music played.**
6. Wait for a while, the enchanting process terminates (which takes much longer than expected), and you will find that the knowledge gem is empty. The bow is still not enchanted.

**Expected behavior**
The enchanter starts playing music. After a few time, the process terminates and the item is enchanted.

**Extra Tests**
The field "crafting_time" looks weired as it always resets.

If you place 3 Crafting Accelerator T3, the enchanting process can be completed without music played. But with 2- accelerators, the process wont terminate until the knowledge gem is empty, and the item is not enchanted.

Enchantment books still can be crafted and upgraded (Process completes normally and music is played). But gilded book cannot copy enchantments from enchanted items, and book placed in center cannot copy enchantments from other enchantment books.

**Minecraft version**
1.21.1, Fabric Loader 0.16.14

**Mod version**
Both observered in 1.9.2 and 1.9.5.
Mod List:
- Architectury 13.0.8
- Cloth Config V15 15.0.140
- Fabric API 0.116.1+1.21.1
- Modmenu 11.0.3
- Modonomicon <Unspecified>
- Revelationary 1.3.9+1.21.1
- Roughly Enough Items 16.0.799
- Roughly Enough Professions 4.0.3 (Reproduced on another modpack without this)
- Spectrum 1.9.5
- Trinkets 3.10.0

**Screenshots**
None.

**Crash Report**
No crash.

Mosklia:
Fixed #764.

If the `currentRecipe` field of an `EnchanterBlockEntity` is `null`, it does not means the enchanter has no valid receipt due to hard-coded center enchanting has no recipe.

### DIFF
--- a/src/main/java/de/dafuqs/spectrum/blocks/enchanter/EnchanterBlockEntity.java
+++ b/src/main/java/de/dafuqs/spectrum/blocks/enchanter/EnchanterBlockEntity.java
@@ -304,7 +304,9 @@ public class EnchanterBlockEntity extends InWorldInteractionBlockEntity implemen
 			return false;
 		}
 		if (enchanter.currentRecipe == null) {
-			return false;
+			// Hard-coded center enchanting recipe has no serialized recipe, so currentRecipe == null
+			// does not mean the enchanter has no valid recipe.
+			return isValidCenterEnchantingSetup(enchanter);
 		}
 		
 		var recipe = enchanter.currentRecipe.value();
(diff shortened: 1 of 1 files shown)


DaFuqs:
Thank you!

Down the line that one should probably get a recipe type + recipe file, which would mean we can ditch that special handling + give modpack makers the option to disable that recipe / tweak costs and more.

PR #766, opened 2025-09-13
Fixed Nightdew cannot grow after broken by Block Breaker

Mosklia:
Nightdews, or `TriStateVineBlock`s, set the block state of the block above it in Block.destroy(), to make the vine can continue to grow. But this method is not invoked after being broken by Block Breaker, making the vine cannot grow. This PR should fix this.

However, it is still recommended to refactor `TriStateVineBlock` class, as Block.destroy() is not always called after a block is destroyed. For example, after an explosion, the lowest vine block's `life_stage` is still `stalk`, making the vine cannot grow after surviving the explosion.


DaFuqs:
Thank you! I agree the whole thing is a mess and seriously needs a pass

PR #774, opened 2025-10-15
Fix accumulation of 'underway' items.

Aquaeyes:
Decrement 'underway' by the amount of items spilt when unable to be inserted. This otherwise prevents sending items when the total amount to be sent is less than the underway amount.


DaFuqs:
ty!

PR #780, opened 2025-11-22
Update en_us.json to mention putting inventory below pedestal

GardenerAether:
This change is just to clarify to players that placing an inventory below the pigment pedestal will deposit items there, so hopefully people have a better understanding of crafting tablet setups ^^

I can work with my friends to get the other locales later but I thought I'd PR this for now.


DaFuqs:
Thank you!

I expanded the explanations a bit further to include how pipes and similar will interact with the Pedestal.

GardenerAether:
Thank you! And yeah this is definitely better than what I wrote ^^

PR #781, opened 2025-11-24
When Mutandis rolling becomes a game of Roulette 

Shibva:
- Added (by hand) all of Spectrums' small flowers and saplings to both mutandis Blacklisted (Except for the Doombloom)
 - Doombloom was excluded by user suggestion. (I guess we do a bit of trolling; Problem?)
 
Fixes issue regarding Enchantment Mutandis referencing #minecraft:small_flowers and #minecraft:saplings; resulting in Spectrum stuff in those to occur (including the Omnious Sapling). Left in Doombloom for comical reasons (hope you're not trying to get Rowan and Alder trees while near your valuables :3 )


DaFuqs:
Thank you!

PR #783, opened 2025-11-25
Block tagspace migration. Tag Unification 1

Shibva:
created/copied existing block files from `c:/` and `minecraft` to serve as a unified location to add new content planned to be added in the future by spectrum; partialy eliminating the need to go back and forth to different files when adding new stuff

there is likely more that can be done but this is all I can muster at the moment. 

Altered all the block tags within `minecraft` and `c/:` to use the pasted files in `spectrum` to hook on to

Overall Change: eliminates the need to swap between `minecraft` and `c:/` namespaces in data files to add new content

Recomendation: follow simular concept when adding compat to mininigate having to go all over the place to add it to a new block (seems like its monstly done so well done; still good idea to keep in mind though). Example: exclude it from mutandis if a new flower is added; progression skip items, etc.

Migrated changes to spectrum data folder to note:

- slabs
- stairs
- walls
- stone_buttons
- flower_pots (refered to as `spectrum:pottable_plants`)
- flowers
- small_flowers
- dirt (refer to as `spectrum:dirt_blocks` to make distint)
- logs_that_burn (referted to as `spectrum:fuel_logs`)
- saplings (Removed Ominous Sappling to prevent it being obtained though tag shenatigans till implimented)
- leaves
- nylium
- striped_logs
- striped _woods
- planks
- wooden_stairs
- fences
- fence_gates
- wooden_doors
- wooden_trapdoors
- wooden_buttons


DaFuqs (approved):
💛

PR #787, opened 2025-12-05
Rest Well, Azzy

Shibva:
- Implimented a check for memories with a `hideTooltip` tag. the existence of this tag will prevent the extra tooltip stuff from appearing on memories
  - This includes ones with mobs, adjust the time and you got yourself a mod-colored custom egg

- Implemented the Tribute for Azzypaaras using the implemented method above; recreated using display Lore properties and the fact that these items retain data when broken. (may this give you an idea for future contributer stuff :P, this one though is special)
   - I got spoilered for this, but it was worth it. Rest well Azzy, and thank you for being a driving force for this mod : )  - Shibva

- Fixed missing imports for the pastel network (how did you miss that, dafuq man (get it?))


Shibva:
Note: the method used is only to substitue for the lack of player memories  in 20.1 (and the fact that I am not even remotly skilled todo that) if your gonna change stuff in the future then remember to check back here.





Rest in Peace Azzypaaras

DaFuqs (approved):
Made a tiny change to the tooltip code.

Tested, seems perfect. Thank you!

Shibva:
Oh geez, I wish I would've thought of that. That would've been much simpler.

Doh!

PR #789, opened 2025-12-06
Fix 'ghost networks' and 'ghost connections'.

Aquaeyes:
Move the code for network split self-removal into a separate method to de-duplicate code as the second call in `checkForNetworkSplit` wasn't removing the network.

Additionally, the path cache wasn't being invalidated when all nodes were still connected.


DaFuqs:
Thank you lots!

DaFuqs (approved):

PR #790, opened 2025-12-08
Backport the Divinity fix to 1.20.1

exaskye:
<b>Una</b> commented:

This is the last custom patch we're running in our modpack after the various unilock and aquaeyes merges, figured I may as well offer it to upstream.


DaFuqs:
Thank you lots!

PR #791, opened 2025-12-10
Project E WoFT nerfing - say NO to extra ticks

Shibva:
This pull request is to serve as the basis for implimenting interactings between spectrum and ProjectE

in particular, the changes made are ment to prevent certain block_entities from having extra ticks being provided via the use the item Watch of Flowing Time - which is what I refer to as an "Tick-Adding Accelerator" on the basis of how it works.

Currently I only added the budding block, but if more are to be added then please suggest and ill do so 
(I do recomend not doing ruin though as it is funny to see it spread fast and makes decay away a bit useful when you fuck up like that lol)

These changes are made in the interest of nerfing the effects this mods unities has on spectrum.


Shibva:
Note for DC tracker bot:

Watch of Flowing time uses a blacklist that works in two areas: `blocks` and `block_entity_type`, placing the thing you want to have the WoFT ticking boost excluded you put it in the apropriate one.

DaFuqs:
TODO: push to 1.21.1, too

src/main/resources/data/projecte/tags/block_entity_type/blacklist/time_watch.json
DaFuqs, line 1: Makes sense, so the speed upgrades still have good use

src/main/resources/data/projecte/tags/blocks/blacklist/time_watch.json
DaFuqs, line 1: Since amethyst buds can be tick accelerated, I feel it makes sense that the others can, too.

No idea how difficult it is to get the time watch in Project E, but progressing there and ditching some waiting time with the geodes sounds reasonable. Getting gem shards is more of a time gate and not a progression skip anyways 😄

Shibva: Part of it also has to do with it causing the bud to grow every tick

These watches can be placed on a special pedestal to accelerate and compound the boost

The watches are fairly easy to make if you focus on it. Project E is like that.

I don't touch the buds because I seen it as being a bit of overreach

DaFuqs (approved):
Letsa goooo

PR #795, opened 2025-12-15
Structure noob mistakes

Shibva:
- Fix Bottom4 of the undergrowth manner missing structure voids


Shibva:
- removed duplicate foxes that were dying using a trick to save the uneffected ones.

Note: NoAI flag acts like /tick freeze for entity processes; use this to your advantage to preserve entities in structures when combined with the special dev block. :P

DaFuqs:
🙏

PR #796, opened 2025-12-16
Pastel Node filter optimisation, and NBT filtering fix

Aquaeyes:
Constructs an internal hashmap of filtered Items for Pastel Nodes, to be used when no NBT / tag selection is occurring, with an internal count of how many special-cased filtering items are in the filter, so as to only use the optimised version when no special-casing is required. Should improve filtering performance in most cases - the cost is setting and removing items from a filter that have NBT is slightly more expensive, but the expense is roughly equivalent to the filter being checked *once* for the node for transfer, so it should almost certainly be a performance increase.

Additionally, prevented Lore from being lowercased in NBT filter handling, as it prevented any matches, and fixed damage checks by checking the NBT as an int as well as a number.

### DIFF
--- a/src/main/java/de/dafuqs/spectrum/blocks/pastel_network/nodes/PastelNodeBlockEntity.java
+++ b/src/main/java/de/dafuqs/spectrum/blocks/pastel_network/nodes/PastelNodeBlockEntity.java
@@ -75,6 +75,8 @@ public class PastelNodeBlockEntity extends BlockEntity implements FilterConfigur
 	protected Direction cachedDirection = null;
 
 	private final List<ItemVariant> filterItems;
+	private final HashSet<Item> filterHashset = new HashSet<>(10);
+	private int nbtCheckingFilterItems = 0;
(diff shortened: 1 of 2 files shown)


DaFuqs (approved):
How did that ever work before

DaFuqs:
Great improvement, I like it a lot, thanks!

TODO: run some final tests

src/main/java/de/dafuqs/spectrum/blocks/pastel_network/nodes/PastelNodeBlockEntity.java
```diff
@@ -351,6 +354,11 @@ public void readNbt(NbtCompound nbt) {
 		
 		if (this.getNodeType().usesFilters()) {
 			FilterConfigurable.readFilterNbt(nbt, this.filterItems);
+			this.filterHashset.clear();
```
DaFuqs, line 357: check: do we need to reset `nbtCheckingFilterItems` here?

Aquaeyes: oh good point, absolutely klsdgf

PR #799, opened 2025-12-30
Fix RedstoneInteractionBlock automatic placement

Aquaeyes:
Additionally, Block Placers now place blocks relative to the direction they themselves are placed in, as opposed to a static one.

Look direction is opposite to the block placer, as by convention and by implementation in Spectrum, directional blocks face towards the player. This would make the use of a Block Placer moot for things such as dispensers, furnaces, or pistons, as they would always face towards the block that placed them, leaving the face covered.

This technically now also allows for block breakers to be placed further inwards to a tree, allowing for, given some advanced and convoluted redstone, all leaves of a small tree to be broken by a player-simulating block breaker using only Spectrum.


DaFuqs:
Awesome change, thank you!

I will make sure to pick it over to 1.21.1, too

PR #800, opened 2026-01-01
Feature: 1.20.1 aria for painters UUID inheritance

Aquaeyes:
**Includes the bugfix also, so probably merge after the bugfix is, if necessary**

Block Placers now remember the player that placed them, and an item tag is added for BlockItems that should be placed as the player that placed the Placer - only the Placer and Breaker are in this tag by default. This means Breakers placed by Placers can break blocks the player that placed the Placer can break.


DaFuqs (approved):

DaFuqs:
I can see this ending up in some fun tomfoolery >:D

PR #801, opened 2026-01-01
tag filtering for black hole chests & negative tag filtering for pastel nodes

Aquaeyes:
the tag filtering update isnt as performant as it could be, but it only runs when the filter changes and takes the same as one pastel node transfer tick with equivalent number of tags so it should be fine


DaFuqs (approved):

DaFuqs:
thank you!
I'll make sure to update the corresponding guidebook page

PR #804, opened 2026-01-13
Fix Rock Candy for Connector

Shibva:
Issue was caused by mixin not doing well with connector transmog; this fix redoes the method to check for it while still allowing for the same end-goal and handling to be accounted for

- commented out submersion check, it was the only one using it; remove if you think its nessisary to purge

- Implitmented Connector-fiendly check, taking advantage of fluid block height being a thing. This should work the same as before while preventing situations like it being on a fence from causing problems (Thanks @800020h for pointing it out)

might not be the most performant, but I dont see it as a problem if its not ticking constantly; **any further impovements made should be checked to make sure it works with connector, otherwise this entire thing was pointless**

Sidenote: Give me a rundown how to sync stuff with changes, I had to wipe my workspace becasue I had old commits that were already pushed, AUGH

### DIFF
--- a/src/main/java/de/dafuqs/spectrum/mixin/EntityApplyFluidsMixin.java
+++ b/src/main/java/de/dafuqs/spectrum/mixin/EntityApplyFluidsMixin.java
@@ -29,7 +29,7 @@ public abstract class EntityApplyFluidsMixin implements TouchingWaterAware {
 	@Override
 	public void spectrum$setActuallyTouchingWater(boolean actuallyTouchingWater) { this.actuallyTouchingWater = actuallyTouchingWater; }
 	
-	@Deprecated //Mixin breaks in Connector Environment -Shibva
+	//This breaks in a connector enviorment for some reason. -Shibva
 	//@Inject(method = "isSubmergedIn", at = @At("RETURN"), cancellable = true)
 	//public void spectrum$isSubmergedIn(TagKey<Fluid> fluidTag, CallbackInfoReturnable<Boolean> cir) {
 	//	if (!cir.getReturnValue() && fluidTag == FluidTags.WATER) {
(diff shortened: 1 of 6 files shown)


src/main/java/de/dafuqs/spectrum/mixin/EntityApplyFluidsMixin.java
```diff
@@ -29,12 +29,13 @@ public abstract class EntityApplyFluidsMixin implements TouchingWaterAware {
 	@Override
 	public void spectrum$setActuallyTouchingWater(boolean actuallyTouchingWater) { this.actuallyTouchingWater = actuallyTouchingWater; }
 	
-	@Inject(method = "isSubmergedIn", at = @At("RETURN"), cancellable = true)
-	public void spectrum$isSubmergedIn(TagKey<Fluid> fluidTag, CallbackInfoReturnable<Boolean> cir) {
-		if (!cir.getReturnValue() && fluidTag == FluidTags.WATER) {
-			cir.setReturnValue(this.submergedFluidTag.contains(SpectrumFluidTags.SWIMMABLE_FLUID));
-		}
-	}
+	@Deprecated //Mixin breaks in Connector Environment -Shibva
```
vgskye: this doesn't mean what you think it means

imreallybadatnames: yeah that's really weird

Shibva: Wasnt sure if I should remove it alltogether tbh, it dosent work in connector for some reason and its unessisary.

I was unsure if I should remove it even though its the only one that was using it

DaFuqs:
> Sidenote: Give me a rundown how to sync stuff with changes, I had to wipe my workspace becasue I had old commits that were already pushed, AUGH

When viewing your github fork, there is an option "sync fork". Once the changes have been synced to your repository, you can "update project" in IDEA to pull those changes to your local project at your PC.

There are other options, but that's the most straightforward one.

PR #805, opened 2026-01-14
Make crystal apothecary behave better with mixed budding types

vgskye:
this PR rewrites the insertion part of the compensation algorithm to insert the compensation reward stacks slot-by-slot weighted-randomly, which results in mixed budding types having (roughly) correct distribution instead of hashes-first-gets-all. while this does make the algorithm more complex and probably slower, it probably doesn't matter as this isn't a very hot code path.

because, well, this is what happens currently if you let crystal apothecary's compensation run for a sufficiently long time with mixed types of budding:
image


DaFuqs:
Very much a needed change, thank you!

Performance isn't that much of a concern, since the logic runs only very rarely. The code is a tad harder to maintain, but the benefits outweigh the cost by a mile here.

PR #807, opened 2026-01-15
Update the life stage of upper vines in onRemove instead of destroy

Auroali:
fixes non-player block breaking locking TriStateVineBlock-based blocks into the mature stage


DaFuqs:
Thank you!

PR #812, opened 2026-01-30
Fix for network splits and fabrication chests.

Aquaeyes:
Fabrication chests are currently partially nonfunctional for recipes that may have a remainder.

### DIFF
--- a/src/main/java/de/dafuqs/spectrum/blocks/chests/RestockingChestBlockEntity.java
+++ b/src/main/java/de/dafuqs/spectrum/blocks/chests/RestockingChestBlockEntity.java
@@ -226,16 +226,14 @@ public class RestockingChestBlockEntity extends SpectrumChestBlockEntity impleme
 	}
 
 	private boolean isRecipeCraftable(Recipe<?> recipe, int index) {
-		 // Don't check if a recipe can be crafted when simulating insertion of its remainders.
-		if(Transaction.isOpen()) {
-			return true;
-		}
-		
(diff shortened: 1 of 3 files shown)


src/main/java/de/dafuqs/spectrum/blocks/chests/RestockingChestBlockEntity.java
```diff
@@ -218,12 +219,18 @@ private boolean tryCraft(RestockingChestBlockEntity chest, int index) {
...
-
+	
+	@SuppressWarnings({"UnstableApiUsage", "Deprecation"})
 	private static boolean isRecipeValid(Recipe<?> recipe) {
 		return recipe instanceof ShapelessRecipe || recipe instanceof ShapedRecipe;
 	}
 
 	private boolean isRecipeCraftable(Recipe<?> recipe, int index) {
+		 // Don't check if a recipe can be crafted when simulating insertion of its remainders.
+		if(Transaction.isOpen()) {
```
imreallybadatnames: nit: for consistency, put a space between the if and condition here
also, the body is so small that it might as well not use curly braces (or even a new line, since it's overall small)

Aquaeyes: hope that works? also moved it below the check for if the resources are present, just in case. it shouldnt be relevant, but may as well

DaFuqs:
Thank you!

PR #821, opened 2026-02-20
What to clone enchant x? TOO BAD, BLACKLIST-EEGEE TIME!

Shibva:
- Added a new Enchantment tag and blacklist check for the Enchanter; Enchantment ids in the `enchanter_blacklisted_enchantments` tag will now be skipped for both enchanting processes. Will still work if the enchantment is on the central item already though. This will enable modpack makers and developers to restrict enchantments that are acquired though other means exclusively to their method.


src/main/java/de/dafuqs/spectrum/registries/SpectrumEnchantmentTags.java
```diff
@@ -12,6 +12,8 @@ public class SpectrumEnchantmentTags {
 	
 	public static final TagKey<Enchantment> SPECTRUM_ENCHANTMENT = of("enchantments");
 	
+	public static final TagKey<Enchantment> ENCHANTER_BLACKLIST = of("enchanter_blacklisted_enchantments");
```
DaFuqs: nit: I am for calling it `enchanter_copying_and_applying_blacklisted`, so the name tells more precisely what it does.

Also might be nice we ship an empty json for that tag, so people can find it a bit easier, if they just look at the data dir

src/main/java/de/dafuqs/spectrum/helpers/SpectrumEnchantmentHelper.java
```diff
@@ -148,7 +148,12 @@ public static Map<Enchantment, Integer> collectHighestEnchantments(List<ItemStac
 			Map<Enchantment, Integer> itemStackEnchantments = EnchantmentHelper.get(itemStack);
 			for (Enchantment enchantment : itemStackEnchantments.keySet()) {
 				int level = itemStackEnchantments.get(enchantment);
-				if (enchantmentLevelMap.containsKey(enchantment)) {
+				
+				if (SpectrumEnchantmentTags.isIn(SpectrumEnchantmentTags.ENCHANTER_BLACKLIST, enchantment)) {
```
DaFuqs: since that method now checks for an enchanter specific it should be moved to the enchanter block entity class, so we don't accidentally use it anywhere else where that exclusion tag should not be applied

DaFuqs (changes requested):

src/main/java/de/dafuqs/spectrum/registries/SpectrumEnchantmentTags.java
Shibva: Agreed

src/main/java/de/dafuqs/spectrum/helpers/SpectrumEnchantmentHelper.java
Shibva: To do this, I would need to go over the list again to sift them out. This could add some performance delay (though I'm not sure) 

The WS also indicates that this is the only line that is used, so here's what I'm gonna do

I'm going to modify this event to include an optional parameter that will act as a blacklist parameter

Shibva: I have opted to make a duplicate of this event and refit it for general use.

The second version of this method includes a second parameter, `blacklist`, which will do this check based on the `KeyTag<Enchantment>` provided

I decided to do it this way so there's a way to use either the filtered or filterless version of this method and refit it if this is to be used in the future more.

If there's a better way to do this, then I'll leave it up to you or someone else with more knowledge on how to do this; the reason I chose this way is both due to the lack of experience/knowledge of Java and that I'm worried filtering it again within the Enchanter Entity might add some performance dela,y even if miniscule

Shibva:
closing this to retry, I think i messed up the branch with the merge lmao

PR #823, opened 2026-02-24
Update zh_cn

SmallYii:


SmallYii:
Hi there,  @DaFuqs 
I noticed that the Windows build for my PR #823 (Update zh_cn) is failing with a 400 Bad Request error. The log shows it can't download the emi-fabric-1.1.22+1.21.1.jar dependency from https://maven.terraformersmc.com.
This seems to be an issue with the upstream Maven repository, not related to the Chinese language files I've submitted. The Ubuntu build is passing successfully.
I don't seem to have the permission to re-run the failed job myself. Could you please help trigger a re-run of the build (21, windows-latest) job? If the issue persists, it might be worth checking the dependency source configuration.
Thanks for your help!

Noaaan:
Don't worry if it happens again. We don't really run any tests, and some of the mod mavens going down happens a lot

PR #824, opened 2026-02-27
What do you mean its opaque, ITS GLASS!!!

Shibva:
- added registry reference for `c:glass_blocks` (common registry) to SpectrumBlockTags

- modified how `FusionShrineBlock` checks for sky obstructions; it will now let any block under `c:glass_blocks` as a valid block REGARDLESS of if its opaque level. This is due to tinted and colored glass returning an opaque value of 15, even though YOU CAN SEE THOUGH IT. This also expands on other glass blocks that can be compatible with any shrine-related builds.
  **This should fix issues with colored and tinted glass blocks by default**
  - Ender glass working is unlikely if it wasn't working already, and is gonna stay that way unless there's a high demand for it, which in that case will require a special check for it.

- TODO: implement in 21.1 fabric and Neo (use `forge:glass` instead)

Tested-by: Shibva <38770780+shibva@users.noreply.github.com>


Shibva:
@DaFuqs not sure why this aint building, it did on my system

DaFuqs:
The current checks match the beacon requirements exactly. It would be kinda weird and annoying to maintain to start making exceptions for various things.

PR #839, opened 2026-03-20
[1.20.1] add surveillance maps

vanutp:
Just a cherry-pick of 695e19766dfe7ba3265290bc7b0102253ee98209. Would be very nice to have this in the 1.20.1 version


DaFuqs:
That's an easy one. Sure thing, I will include it in the next version

DaFuqs (approved):

PR #855, opened 2026-04-06
Fix boundary Ruin portal behavior for bedrock drops and Nether roof

mikaba-eu:
## Summary

This PR fixes two edge cases around Deeper Down portal creation from boundary Ruin blocks.

## Problem

When a Ruin block at the world boundary is broken, the current logic places a Deeper Down portal during block removal. In flat or single-layer bedrock setups, this means converted bedrock at the Overworld floor is immediately replaced by a portal, which prevents the expected Shattered Bedrock (`bedrock_dust`) drop from being obtained normally.

The same boundary logic also creates a Deeper Down portal on the Nether roof, even though that portal is not useful there.

## Changes

- Move boundary portal placement from `onRemove` to `playerDestroy`.
- Let the normal player break flow finish before placing the portal block, so loot drops are preserved.
- Keep valid boundary portal creation for the Overworld floor and other intended cases.
- Disable Deeper Down portal creation on the Nether roof by removing the Nether ceiling condition.

## Why

This keeps the intended progression intact for modpacks with custom bedrock generation while also avoiding non-functional portal generation on the Nether roof.

## Testing

- `./gradlew compileJava`
- `./gradlew build`

### DIFF
--- a/src/main/java/de/dafuqs/spectrum/blocks/decay/RuinBlock.java
+++ b/src/main/java/de/dafuqs/spectrum/blocks/decay/RuinBlock.java
@@ -83,23 +83,20 @@ public class RuinBlock extends DecayBlock {
 	}
 	
 	@Override
-	public void onRemove(BlockState state, Level level, BlockPos pos, BlockState newState, boolean moved) {
-		super.onRemove(state, level, pos, newState, moved);
+	public void playerDestroy(Level level, Player player, BlockPos pos, BlockState state, @Nullable net.minecraft.world.level.block.entity.BlockEntity blockEntity, ItemStack tool) {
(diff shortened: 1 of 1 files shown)


imreallybadatnames:
> Disable Deeper Down portal creation on the Nether roof by removing the Nether ceiling condition.

Why though? This is intentional IIRC.

mikaba-eu:
> > Disable Deeper Down portal creation on the Nether roof by removing the Nether ceiling condition.
> 
> Why though? This is intentional IIRC.

Didn't know that. But the portal don't work in the nether either. At least if u try to enter it from above.

DaFuqs:
The change for the drop to process definitely is a great change, thank you!

But yes, the portal forming in the Nether roof is a feature. Using it will loop around to the Nether floor and vice versa

PR #856, opened 2026-04-06
"Quite of obvious" -> "Quite obvious"

SubordinalBlue:
Very small edit; just caught my eye while playing


imreallybadatnames:
LGTM

PR #883, opened 2026-05-14
Hide revealations from jade

justliliandev:
Hi,

a friend had a problem with spectrums hidden blocks being visible when you look at them with the jade overlay.


justliliandev:
Moved this over to the Revealations repo: https://github.com/DaFuqs/Revelationary/pull/32
I should have checked if this repo was the correct place

PR #894, opened 2026-06-03
Fixes 2 dupe glitches involving presents, the Pride wrapping paper being unobtainable, and a quick-crafting fix.

myriantics:
## Problems
- Currently, ```WrapPresentRecipe```s copy the input present stack into the output slot - including its count. This enables a duplication glitch involving input present stacks of size greater than 1.
- ```PresentBlock```s currently have a similar issue; copying the entire stack regardless of count. This enables a duplication glitch triggered by placing a present from a stack of size greater than 1 - breaking the present will yield the stack that was originally used to place it, including its count, even though only 1 present was consumed to place it.
- Using a Spore Blossom as the wrapping paper for a present should grant the ```PRIDE``` wrapping paper - however it currently grants ```STRIPED``` due to Spore Blossoms being in the ```#minecraft:flowers``` tag, which is checked first.
- Wrapping a present via quick-crafting from within the player's inventory causes its ```PROFILE``` component to not be populated.

## Changes
- ```WrapPresentRecipe``` has been tweaked to only copy the input present stack with a count of 1, fixing the duplication glitch.
- ```PresentBlock``` now only copies the stack used to place it with a count of 1, fixing that duplication glitch.
- The check for the ```STRIPED``` WrappingPaper has been moved to the bottom of the if/else chain in ```WrapPresentRecipe.getPresentForStack()``` as it's the most broad.
- Hooked a mixin into ```InventoryMenu.moveItemStackTo()``` that calls ```Item.onCraftedBy()``` before the quick movement operation is run, allowing the changes to be applied properly. This mixin checks if the output item is an instance of ```PresentBlockItem``` before calling the method to preserve behavior of other items.

## Testing
- Tested crafting presents in multiple crafting menus with multiple stack sizes - results are now as expected.
- Tested placing presents and picking them back up with multiple stack sizes - no duplicate presents were produced.
- Tested using a Spore Blossom as a wrapping paper when crafting a present - now outputs the proper wrapping paper.
- Wrapped a present via quick-crafting in the player's inventory crafting grid - ```PROFILE``` component is updated as expected.


DaFuqs:
Good catch with the wrapping paper!

How did you get presents with content to stack? They should not be able to accept items when having a count >1 or be able to stack at all.

myriantics:
I got presents with content to stack by splitting them into single stacks, splitting the item I wanted to dupe into equal stacks, putting the content in each present one by one, then stacking them.
However, these dupes would still work even if that was not possible, albeit to a lesser degree, because you can wrap presents with no contents, enabling you to dupe infinite empty, wrapped presents.

DaFuqs:
Changing the presents stack size to 1 should mitigate this issue partially: https://github.com/DaFuqs/Spectrum/commit/cc3bfb7022f01c89423f2f3cb0a23216301b20bc

I will see if that allows us to skip the mixin.

DaFuqs:
That seems to work nicely!

Can you review / rest again, please?

myriantics:
The issue with the player's profile not being assigned when quick-crafting in specifically the player's inventory crafting menu reappears without the mixin.

This is because ```AbstractContainerMenu.moveItemStackTo()```, on line 664, splits the stack it's given when inserting an item into a slot because the same logic is used for both stackables and unstackables. This causes the stack reference passed to ```Slot.onQuickCraft()``` > ```ResultSlot.checkTakeAchievements``` > ```Item.onCraftedBy()``` to be outdated, as the copy has already been put into the slot when called.

Vanilla fixes this in ```CraftingMenu``` on line 137 by calling ```Item.onCraftedBy()``` before the quick movement logic is called. This fix was likely implemented to fix a similar bug caused with Maps as they also use this method, but was not propagated to ```InventoryMenu``` because ```MapExtensionRecipe``` requires the full-size 3x3 crafting grid.

Ultimately this is a small issue, and I understand not wanting to have a whole mixin for a super-edge case that is not very harmful such as this - it's just an issue I encountered while testing the other bugs.

Demo video of the bug:
[https://www.youtube.com/watch?v=u8oKleeaJKU](url)

DaFuqs:
Props to you for the detailed writeup!

It is just a small issue really, but still nice to have working properly.

PR #896, opened 2026-06-03
Fabric 1.21.1 - Fix LC fluid interaction dupe, fix swimming

SaphireLattice:
The fluid collision thing feels clunky but it should work so uh, shrug?

The mixin could have been merged into the other one to reduce confusion but this works as a quick fix.

I've tested the mixin and it fixes it all fine, but can't test the fluid interaction thing as the API's down and `runClient` just hangs.


DaFuqs:
thank you!

PR #906, opened 2026-06-14
Fix return value of SingleInk storage, so it stops duping initial fill

SaphireLattice:
It's extra noticeable if there are any addons that let you transfer a lot of ink at once into an empty storage

Was initially noticed in https://github.com/vgskye/appspec/pull/7 but I forgor


imreallybadatnames (approved):
yummers

PR #907, opened 2026-06-14
Make singleink storage check ink acceptance more

SaphireLattice:
Lets us simplify the subsclasses some.

Also makes sure that some addon doesn't accidentally push wrong color of ink into an item that shouldn't accept any, if they skip `accepts`, as `addEnergy` can generally be called with wrong colors either way.


imreallybadatnames (approved):
sure
do note however that this is likely to blow up when the ink refactor is merged from the neoforge branch

PR #909, opened 2026-06-15
Don't wirelessly link an ItemStack's component to a live player

SaphireLattice:
Per neoforge docs (but it's a minecraft's own class, they just documented some about it):
> `MobEffectInstances` are mutable. If you need a copy, call `new MobEffectInstance(oldInstance)`.

Stops the item stack from having an effect updated by the player's effect ticking down. Or whatever might be going on in the depths of minecraft codebase.

Also behaved really oddly (crash potentially?) when you tried to store it in some mods' special item containers


imreallybadatnames (approved):
satisfactory

PR #918, opened 2026-07-04
Fix wrong check in Neoforge 1.21.1 particle payload

kawaiixissa13:
i havent tested this but it matches what the fabric version does: https://github.com/DaFuqs/Spectrum/blob/0d1f13e931a00244c2fd245f1343ff1d3d4af8ca/src/main/java/de/dafuqs/spectrum/networking/s2c_payloads/PlayParticleAroundBlockSidesPayload.java#L32


DaFuqs:
Thank you!

PR #927, opened 2026-07-18
A small fix for farmers delight cutting compat, and adds a new cutting recipe for noxcaps

Vihri:
Fixes an issue with gemstone cutting recipes not accepting multitools and the malachite staff in gemstone block splitting recipes.

Also adds cutting recipes to noxcaps for making myceylon and stripped noxcaps, but there could also be a good reason for this to not exist that i dont know about.


DaFuqs:
Thank you!

The Myceylon addition is fine. It is mostly a puzzle how to get these which REI/EMI will spoil here, but it is nothing but a minor discovery anyways.

DaFuqs (approved):

Issue #932, opened 2026-08-01
[BUG] Item slot decorations draw OVER hotbar items when ImmediatelyFast has hud batching enabled


kawaiixissa13:
**Describe the bug**
with hud batching on in immediatelyfast, item decorations are drawn over items even when transparent which means the depth stencil ends up blocking it

**To Reproduce**
1. have an item with the SlotEffect.PULSE decoration
2. look at it in the hotbar with ImmediatelyFast hud batching enabled (on by default w/ the mod)

**Expected behavior**
draws normally

**Version information**
Spectrum Version: 1.10.5
Minecraft Version: 1.21.1
Mod Loader: Fabric

**Screenshots**

<img width="494" height="98" alt="Image" src="https://github.com/user-attachments/assets/d08cedd7-2f9b-4991-b9a3-c37628d4112a" />

item draw call with depth test overlay:
<img width="460" height="85" alt="Image" src="https://github.com/user-attachments/assets/38f6616b-5afa-4791-910c-87395ab23fdf" />

depth buffer before item draw call:

<img width="689" height="92" alt="Image" src="https://github.com/user-attachments/assets/fb72ddbc-3b0c-4865-b19f-4ec6d62b285f" />

depth buffer after item draw call:
<img width="542" height="84" alt="Image" src="https://github.com/user-attachments/assets/0f587611-04a1-4c1b-881d-4ffb204eabd4" />

(the item sprite are at like depth 0.4725 and the gradient decorations are at like 0.47, u can see it eve naffects the ones that are just borders but it only hides one pixel so its harder 2 notice)

you can see that its happening with any of the items that draw a background gradient (SlotEffect.PULSE in spectrum's code), and i think its because of the depth test, it doesnt rly handle transparency very well. is it rly necessary to have the depth test on for this, it only saves like a few pixels of overdraw doesnt it? and its not on for the inventory

also it doesnt happen in the inventory, only in the hotbar
<img width="874" height="371" alt="Image" src="https://github.com/user-attachments/assets/4809c2f7-6231-4110-84d6-182c11fdf4d1" />

i think what amkes the difference bt the 2 is that `AbstractContainerScreen.render` does `RenderSystem.disableDepthTest();` while `Gui.render` does `RenderSystem.enableDepthTest();` and it isnt disabled in `Gui.renderItemHotbar`. idk why it only happens with hud batching tho maybe its like an ordering thing

[...]

PR #936, opened 2026-08-02
Change injection point for renderItemDecorations method to be compatible with ImmediatelyFast

### LINKED ISSUE #558: [Fabric 1.21.1] Hud batching makes transparent Spectrum item slot decorations cull item sprites via depth test
minecraft version 1.21.1
fabric loader version 0.19.3
fabric api version 0.116.15+1.21.1
spectrum version 1.10.5
immediatelyfast version 1.6.11+1.21.1

i think its probably related to this commit https://github.com/RaphiMC/ImmediatelyFast/commit/4f81ea0c0e1ea1c392bec56e9720276f08f7ebd5 but i could be wrong on that

Image

item draw call with depth test overlay:
Image

depth buffer before item draw call:

Image

depth buffer after item draw call:
Image

(the item sprite are at like depth 0.4725 and the gradient decorations are at like 0.47, u can see it eve naffects the ones that are just borders but it only hides one pixel so its harder 2 notice)

you can see that its happening with any of the items that draw a background gradient (SlotEffect.PULSE in spectrum's code), and i think its because of the depth test, it doesnt rly handle transparency very well. is it rly necessary to have the depth test on for this, it only saves like a few pixels of overdraw doesnt it? and its not on for the inventory

also it doesnt happen in the inventory, only in the hotbar
Image

i think what amkes the difference bt the 2 is that `AbstractContainerScreen.render` does `RenderSystem.disableDepthTest();` while `Gui.render` does `RenderSystem.enableDepthTest();` and it isnt disabled in `Gui.renderItemHotbar`. idk why it only happens with hud batching tho maybe its like an ordering thing

RaphiMC:
Fixes https://github.com/RaphiMC/ImmediatelyFast/issues/558 and https://github.com/DaFuqs/Spectrum/issues/932

### DIFF
--- a/src/main/java/de/dafuqs/spectrum/mixin/client/DrawContextMixin.java
+++ b/src/main/java/de/dafuqs/spectrum/mixin/client/DrawContextMixin.java
@@ -24,7 +24,7 @@ public abstract class DrawContextMixin {
 	@Shadow
 	public abstract void renderOutline(int x, int y, int width, int height, int color);
 	
-	@Inject(method = "renderItemDecorations(Lnet/minecraft/client/gui/Font;Lnet/minecraft/world/item/ItemStack;IILjava/lang/String;)V", at = @At(value = "INVOKE", target = "net/minecraft/world/item/ItemStack.getCount ()I", ordinal = 0))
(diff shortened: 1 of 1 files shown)


DaFuqs:
Thank you lots!

PR #944, opened 2026-08-12
Update zh_cn.json

SmallYii:
pls merge prior to release to avoid frequent pull requests.


DaFuqs:
Thank you, as always!

